PR #1, opened 2018-11-13
Add value type

xichen2020:
cc @notbdu

This PR adds the value type alongside object type to represent different types of values.


xichen2020:
Yes that'll be addressed in a followup PR.

PR #5, opened 2018-11-19
Add iface def for int encoder.

notbdu:
Adds interface definition for int encoder/decoder.

Adds tests to show how the interface is to be used.


encoding/int/encoding.go
```diff
@@ -0,0 +1,42 @@
+package int
+
+import "io"
+
+// Encoder encodes int values.
+type Encoder interface {
+	// Encode a block if ints.
+	Encode(values []int) error
+	// Bytes reutrns the encoded bytes of the encoder.
```
xichen2020: This is implementation detail and should not be part of interface definition.

Also typo `reutrns` -> `returns`.

encoding/int/encoding.go
```diff
@@ -0,0 +1,42 @@
...
+// Iterator lazily produces int values decoded from a byte stream.
+type Iterator interface {
+	io.Closer
+
+	// Next returns true if there is another value
+	// in the data stream. If it returns false, err should be
+	// checked for errors.
+	Next() bool
+	// Current returns the current value in the iteration.
+	Current() int
```
xichen2020: Do you need to return metadata in here?

encoding/int/generate.go
```diff
@@ -0,0 +1,3 @@
+package int
```
xichen2020: Conventionally we put `generate.go` for mocks in `generated/mocks/`. Perhaps move it there?

For the Makefile, perhaps follow the convention of mock generation in `https://github.com/m3db/m3aggregator/blob/master/Makefile`? Try copying any lines containing `mock` and dependencies and run `make mock-gen` from the top level dir, and you should see the generated mock files in `encoding/int/`

xichen2020:
PR style: please add a brief one to two sentence summary in the PR description.

Also check if the tests are failing. Seems in this case you haven't run `glide update`.

encoding/int/encoding.go
notbdu: I was thinking that metadata should be private to the encoder.

encoding/int/encoding.go
notbdu: Do you want to return bytes in the `Encode` method then?

I thought you mentioned that we should return the encoded bytes in a `Bytes()` call for some reason.

encoding/int/generate.go
notbdu: I'll move it and edit the `Makefile`.

encoding/int/encoding.go
notbdu: Yea, each encoder will prob have diff metadata implementations. I'm thinking it would be fine to leave it opaquely in the encoded byte slice since it _probably_ won't be used outside of the encoder. 

On the other hand... the encoder should return the "Block metadata" which contains meta info about that data block to be used during queries?

encoding/int/encoding.go
notbdu: Will probably just return `Bytes()` as an `Encode()` return value. Having it as a separate fn call looks like a non-obvious API.

encoding/int/encoding.go
```diff
@@ -0,0 +1,42 @@
...
+	// Bytes reutrns the encoded bytes of the encoder.
+	// The format of the byte slice is as follows:
+	//   - version:                 1 byte
+	//   - metadata length bytes:   VInt
+	//   - metadata bytes:          Variable bytes
+	//   - data bytes:              Variable bytes
+	Bytes() []byte
+	// Reset internal memory buffers manually.
+	// Otherwise, buffers are reset upon the next call to `Encode`.
+	Reset()
```
xichen2020: I'd perfer to document that the callers should always reset buffers before calling `Encode`, instead of implicitly relying on the `Encode` API to reset the buffers for the sake of the "least responsibilities` principle.

encoding/int/encoding.go
xichen2020: Hm that feels to me should not be part of encoder's responsibilities, but rather the decoder or a component that's higher level above the decoder to know about the meta information.

What specifically is included in the meta information you are thinking about?

notbdu: Some of the meta information will *only* be used by the decoder to decode the data like you said. That information will not be exposed and will live in the head of the `[]byte` slice returned by the encoder `Bytes()` method.

Some other info will be stored as block metadata for querying purposes. For example:
- cardinality
- min/max
- doc id start/stop
- unix ts start/stop

Also updated my segment file design doc: https://docs.google.com/document/d/1gMsn1VCiZELVm_tDSOy4a9pbvn5e1gXO7-LyN2Ma2Oo/edit

encoding/int/encoding.go
notbdu: 👍

notbdu:
PS - I feel like glide is super flaky... I often run into this:
```
bodu @ bodu-C02WP0EWHTDG eventdb (bdu/encodingiface) $ glide update
[INFO]  Downloading dependencies. Please wait...
[INFO]  --> Fetching updates for github.com/stretchr/testify.
[INFO]  --> Detected semantic version. Setting version for github.com/stretchr/testify to v1.2.2.
[INFO]  Resolving imports
[ERROR] Error scanning github.com/golang/mock/gomock: cannot find package "." in:
        /Users/bodu/.glide-uber/cache/src/https-github.com-golang-mock/gomock
[ERROR] Failed to retrieve a list of dependencies: Error resolving imports
```

notbdu:
Also, `make metalint` and `make test-internal` pass for me locally...

xichen2020:
I'd guess probably it passes locally because you have gomock in your `GOPATH` but not in `vendor`. From your top dir try `rm -rf vendor/ && glide cc && glide update -s -v -u` and see if it helps.

encoding/int/encoding.go
xichen2020: Can you remove the `Otherwise, ...` comment if that no longer applies?

encoding/int/encoding.go
xichen2020: Interesting, what's the distinction between metadata stored as part of the encoded bytes and the metadata stored as block metadata exposed for querying? How do you plan to store the block metadata?

notbdu: Discussed offline. Encoder is responsible for parsing out metadata info from the actual data.

encoding/int/encoding.go
```diff
@@ -0,0 +1,46 @@
+package int
+
+import "io"
+
+// Encoder encodes int values.
+type Encoder interface {
+	// Encode a collection of ints.
+	// Callers should explicitly call `Reset` before
+	// subsequent call to `Encode`.
+	Encode(Iterator) error
```
xichen2020: nit: conventionally we spell out the var names in the APIs for clarity.

encoding/int/encoding.go
```diff
@@ -0,0 +1,46 @@
...
+	// Bytes returns the encoded bytes of the encoder.
+	// The format of the byte slice is as follows:
+	//   - version:                 1 byte
+	//   - metadata length bytes:   VInt
+	//   - metadata bytes:          Variable bytes
+	//   - data bytes:              Variable bytes
+	Bytes() []byte
```
xichen2020: Are the comments still applicable here?

encoding/string/encoding.go
```diff
@@ -0,0 +1,46 @@
+package string
+
+import "io"
+
+// Encoder encodes string values.
+type Encoder interface {
+	// Encode a collection of strings.
+	// Callers should explicitly call `Reset` before
+	// subsequent call to `Encode`.
+	Encode(Iterator) error
```
xichen2020: Same comments as those in the `int/encoding.go`

encoding/string/encoding.go
```diff
@@ -0,0 +1,46 @@
+package string
+
+import "io"
+
+// Encoder encodes string values.
+type Encoder interface {
```
xichen2020: Can you create an issue for templatizing these interfaces if they are exactly the same aside from the type difference?

encoding/int/encoding.go
notbdu: Yes, they are. The encoded bytes still needs to contain metadata info useful for the decoding process.

encoding/int/encoding.go
notbdu: 👍

encoding/int/encoding.go
xichen2020: I'd remove it from the API comments. Feels like an implementation detail to me.

encoding/string/encoding.go
xichen2020: Yeah nw, I'm used to creating issues for things that do not require immediate attention but may be worth revisiting at some point.

encoding/string/encoding.go
xichen2020: Ping, should add var name here.

encoding/string/encoding.go
```diff
@@ -0,0 +1,46 @@
...
+	// Callers should explicitly call `Reset` before
+	// subsequent call to `Encode`.
+	Encode(Iterator) error
+	// Bytes returns the encoded bytes of the encoder.
+	// The format of the byte slice is as follows:
+	//   - version:                 1 byte
+	//   - metadata length bytes:   VInt
+	//   - metadata bytes:          Variable bytes
+	//   - data bytes:              Variable bytes
+	Bytes() []byte
```
xichen2020: Can you remove the encoding format here? Probably implementation detail that shouldn't be part of the API docs.

encoding/string/string.go
```diff
@@ -0,0 +1,29 @@
...
+import "bytes"
+
+type encoder struct {
+	buf *bytes.Buffer
+}
+
+// NewEncoder returns a new encoder instance.
+func NewEncoder() Encoder {
+	return &encoder{
+		buf: bytes.NewBuffer(make([]byte, 0)),
```
xichen2020: Any reason not to just do
```
type encoder struct {
  buf bytes.Buffer
}
```
?

encoding/string/encoding.go
notbdu: https://github.com/xichen2020/eventdb/issues/14

encoding/string/encoding.go
notbdu: 👍

encoding/string/string.go
notbdu: I had considered making the initial buffer size configurable and `bytes.NewBuffer` returns a `*bytes.Buffer` but I'll just use `bytes.Buffer` for now as it doesn't require initialization.

encoding/string/encoding.go
notbdu: 👍

encoding/int/encoding.go
notbdu: 👍

encoding/string/string.go
xichen2020: In that case it's fine to leave it as is, didn't know that was your intention based on the PR.

For that purpose I'd imagine you'd pass in an options for specifying the initial buffer sizes etc.

PR #9, opened 2018-11-22
Support parsing JSON messages up to a maximum depth

xichen2020:
cc @notbdu 

This PR adds logic to support parsing JSON messages up to a configurable maximum depth. The depth limitation currently only applies to JSON objects in the input message. If an object contains nested key/values that are below the maximum depth, the parser simply skips over any value for fast processing, and return an empty object instead. The benchmark shows that for complex objects, parsing with a smaller maximum depth (1 in the benchmarks) can result in a parsing speedup of **2x - 10x.**

For example, for input JSON string below:
```
{
	"foo": 123,
	"bar": [
		{
			"baz": {
				"cat": 456,
				"car": 789
			},
			"dar": ["bbb"]
		},
		666
	],
	"rad": ["usa"],
	"pat": {
		"qat": {
			"xw": {
				"woei": "oiwers",
				"234": "sdflk"
			},
			"bw": 123
		},
		"tab": {
			"enter": "return"
		},
		"bzr": 123
	}
}
```
A parser with a maximum depth of 1 will parse it as
```
{
       "foo": 123,
       "bar": [
           {}, 
           666
       ],
       "rad": ["usa"],
       "pat":{}
}
```
and a parser with a maximum depth of 2 will parse it as
```
{
       "foo": 123,
       "bar": [
             {
                 "baz": {},
                 "dar": ["bbb"]
             },
            666
       ],
       "rad": ["usa"],
       "pat": {
            "qat": {},
            "tab": {},
            "bzr": 123
       }
}
```

### DIFF
--- a/parser/json/parser.go
+++ b/parser/json/parser.go
@@ -102,8 +102,9 @@ type parser struct {
 	maxDepth int
 	cache    cache
 
-	str string
-	pos int
+	str   string
+	pos   int
+	depth int
 }
 
 // NewParser creates a JSON parser.
@@ -120,6 +121,7 @@ func (p *parser) Reset() {
 	p.cache.reset()
 	p.str = ""
 	p.pos = 0
+	p.depth = 0
 }
 
 func (p *parser) Parse(str string) (*value.Value, error) {
@@ -206,6 +208,16 @@ func (p *parser) parseObject() (*value.Value, error) {
 		return emptyObjectValue, nil
 	}
 
+	if p.depth >= p.maxDepth {
+		// Skip parsing values due to crossing maximum depth threshold.
+		if err := p.skipObjectValue(); err != nil {
+			return nil, err
+		}
+		// NB: If the parser successfully skips the object value without
+		// encountering any errors, return an empty object value.
+		return emptyObjectValue, nil
+	}
+
 	var kvs *value.KVArray
 	for {
 		p.skipWS()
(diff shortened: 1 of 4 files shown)


parser/json/parser.go
```diff
@@ -261,6 +275,46 @@ func (p *parser) parseObject() (*value.Value, error) {
 	}
 }
 
+// Precondition: The parser has seen a '{', and expects to see '}'.
+func (p *parser) skipObjectValue() error {
```
notbdu: ++!

parser/json/parser.go
```diff
@@ -229,10 +241,12 @@ func (p *parser) parseObject() (*value.Value, error) {
 
 		// Parse out the value.
 		p.skipWS()
+		p.depth++
 		v, err := p.parseValue()
```
notbdu: I was expecting to see another `p.depth++` when it enters a nested obj inside of `p.parseValue()`.

notbdu:
Left a comment about `p.depth`.

parser/json/parser.go
xichen2020: Hm not sure if I follow, so the maximum depth only applies to JSON objects, in other words the only time `depth` may be incremented is in `parseObject`, which is where `depth` is currently incremented. Specifically, `parseObject` will call `parseValue` to parse out the values for each key, and for values that are objects, `parseValue` will in turn call `parseObject` which increments `depth`.

There's also a unit test `TestParseValueMaxDepth` to test various cases.

notbdu: Woops, the way github collapsed the diff it looked like this was inside of `Parse()` XD.

notbdu:
LGTM!! 🚢

PR #12, opened 2018-11-25
Bootstrap database write path

xichen2020:
cc @notbdu 

This PR adds the initial logic to bootstrap the database write path.

### DIFF
--- a/sharding/shardset.go
+++ b/sharding/shardset.go
@@ -5,11 +5,10 @@ import (
 	"math"
 
 	"github.com/m3db/m3cluster/shard"
-
 	"github.com/spaolacci/murmur3"
 )
 
-// HashGen generates HashFn based on the length of shards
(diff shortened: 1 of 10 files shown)


sharding/shardset.go
```diff
@@ -0,0 +1,164 @@
...
+
+type shardSet struct {
+	shards   []shard.Shard
+	ids      []uint32
+	shardMap map[uint32]shard.Shard
+	fn       HashFn
+}
+
+// NewShardSet creates a new sharding scheme with a set of shards.
+func NewShardSet(shards []shard.Shard, fn HashFn) (ShardSet, error) {
```
notbdu: Is there any way to grow the # of shards?

notbdu: It looks like you'd need to create a new shard set each time?

PR #16, opened 2018-11-28
Segment flushing and persistence part 2

xichen2020:
cc @notbdu 

This PR is the part 2 of the segment flushing and persistence logic. Specifically, it threads the `Flush` call throughout the database ingestion path and fleshes out the persistence related data structures more in preparation for actual persistence.

NB: The type-specific persistence functions now still takes slices as the input parameter, but I can change them to be iterators later on if desired once I finish wiring up the pieces.

### DIFF
--- a/storage/shard.go
+++ b/storage/shard.go
@@ -4,12 +4,20 @@ import (
 	"errors"
 	"sync"
 
+	xerrors "github.com/m3db/m3x/errors"
 	"github.com/xichen2020/eventdb/event"
+	"github.com/xichen2020/eventdb/persist"
 )
 
 type databaseShard interface {
+	// ID returns the shard ID.
+	ID() uint32
+
 	// Write writes an event within the shard.
 	Write(ev event.Event) error
+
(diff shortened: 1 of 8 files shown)


storage/shard.go
```diff
@@ -50,6 +61,41 @@ func (s *dbShard) Write(ev event.Event) error {
 	return segment.Write(ev)
 }
 
+// TODO(xichen): retry logic on segment persistence failure.
+func (s *dbShard) Flush(ps persist.Persister) error {
+	var multiErr xerrors.MultiError
+	s.Lock()
+	activeSegment := s.active
+	s.active = newDatabaseSegment(s.opts)
```
notbdu: During the flush phase, it looks like `activeSegment` is invisible to the shard. We won't be able to query it?

xichen2020: Honestly havent' given much thought to the read API at this point, was thinking either you or I'll make whatever changes necessary when implementing the read API.

PR #20, opened 2018-11-30
Segment flushing and sealing part 3

xichen2020:
cc @notbdu 

This PR adds the part 3 of the segment flushing and sealing logic. This PR uses the encoders to perform the persistence to segment files.


notbdu:
Changing the `Reset()` iface impl to take a slice of values as the data to be used for reset is problematic for the encoders since the logic in the encoder does this:

```
1. Loop over data to generate metadata.
2. Call `Reset()` to reset iteration.
3. Loop over data again to encode.
```

Could the impl be changed to take ` Reset(nil)` which will reset the iteration internally on the same data slice?

PR #22, opened 2018-11-30
Update Encoding Package

notbdu:
- Refactor encoding package. 
- Update encoder/decoder ifaces. 
- Update encoder/decoder iface usage.
- Add temp constructors.


encoding/encoding.go
```diff
@@ -0,0 +1,145 @@
...
+	// Reset should be called between `Encode` calls.
+	Reset()
+}
+
+// StringDecoder decodes int values.
+type StringDecoder interface {
+	// Decode the source bytes.
+	// Callers should explicitly call `Reset` before
+	// subsequent call to `Decode`.
+	Decode(src Reader) (StringIterator, error)
```
xichen2020: I'm not sure the iterator can rewind during decoding though. Also doesn't look like we need that functionality. Thoughts?

xichen2020: Also perhaps we just rename `StringDecoder` to `StringIterator`? Seems all it does it producing an iterator.

encoding/constructor_temp.go
```diff
@@ -0,0 +1,117 @@
...
+package encoding
+
+import "io"
+
+// STRING
+
+type stringEncoder struct{}
+
+// NewStringEncoder returns a new StringEncoder instance.
+func NewStringEncoder() StringEncoder {
```
xichen2020: I think the ctor should take an `io.Writer` no?

encoding/constructor_temp.go
```diff
@@ -0,0 +1,117 @@
...
+	return &stringEncoder{}
+}
+
+func (s *stringEncoder) Encode(values StringIterator, writer io.Writer) error {
```
xichen2020: The writer should be moved to the ctor.

encoding/encoding.go
notbdu: It can't rewind during decoding. Let's just remove `StringDecoder` then. There already is a `StringIterator`.

encoding/constructor_temp.go
notbdu: The `Encode()` call takes to writer. Otherwise we'd need a method to replace the `io.Writer` target.

xichen2020: Yeah I like that better, have a both the ctor take an `io.Writer` and a `Reset(io.Writer)` method. Feels a bit strange IMO that the `Encode` call needs to take an `io.Writer`.

encoding/encoding.go
xichen2020: I meant currently the `StringIterator` has a `Rewind` method in the API due to embedding the `BaseIterator`, so if the iterator can't rewind during decoding, then the string decoder shouldn't return a `StringIterator`.

I was thinking perhaps we name the iterator passed to encoder a `RewindableStringIterator`, and the iterator returned by the decoder `ForwardIterator` taking inspiration from c++ std iterator naming patterns. IOW you'd have
```
type StringEncoder interface {
  Encode(it RewindableIterator) error
}

type ForwardStringIterator interface {
  Next() bool
  Current() string
  Err() error
  Close() error
}

type RewindableString Iterator {
  ForwardStringIterator

  Reset(values []string)
  Rewind()
}
```
Thoughts?

encoding/constructor_temp.go
xichen2020: Actually I think this is fine after giving it a second thought, although I'd probably move the writer to the first param though.

PR #29, opened 2018-12-04
Pooling value arrays on the write path

xichen2020:
cc @notbdu 

This PR adds logic to pool value arrays on the write path.


x/pool/bucketized_bool_array_pool.gen.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	p.buckets = buckets
+	p.alloc = alloc
+}
+
+// Get gets a value from the pool.
+func (p *BucketizedBoolArrayPool) Get(capacity int) []bool {
+	if capacity > p.maxBucketCapacity {
+		p.maxAlloc.Inc(1)
+		return p.alloc(capacity)
```
notbdu: Should we allocate the requested capacity or should we have well defined bucket sizes and go up one size?

e.g.
```
1 << 2
1 << 3
1 << 4
1 << 5
1 << 6
etc.
```

I was thinking about the case where you get multiple requests to allocate increasingly larger slices but only by a small margin resulting in many allocs. Might be too much of an edge case to handle though.

xichen2020: That is indeed the intention, i.e., having bucket sizes that are exponentially increasing by a factor of 2. However this should be configured externally as there may be usecases where this is not desirable. For our usecases yes, a factor of 2 should work just fine.

xichen2020:
Re "a type specific array pool inside a bucket can only allocate arrays of the same size": Correct, the bucketized pool internally has a number of buckets each of which has a predefined capacity, as well as a backing pool with arrays whose capacity match the bucket capacity.

PR #30, opened 2018-12-04
Implement main.

notbdu:
- Adds config impl for main.
- Loads config and instantiates db/service/server.
- Runs server.
- Minor fix to call `Done()` when persister is finished flushing.


notbdu:
LGTM.

PR #31, opened 2018-12-04
Add bulk write endpoint.

notbdu:
- Adds a bulk write endpoint.

### DIFF
--- a/server/http/handlers/service.go
+++ b/server/http/handlers/service.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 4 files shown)


server/http/handlers/service.go
```diff
@@ -146,3 +152,114 @@ func (s *service) Write(w http.ResponseWriter, r *http.Request) {
...
+		start int
+		stop  int
+	)
+	for stop < len(data) {
+		stop = bytes.Index(data[start:], writeBulkDelimiter)
+		if stop == -1 {
+			stop = len(data)
+		}
+
+		v, err := p.ParseBytes(data[start:stop])
```
xichen2020: Hm we should refactor and try to reuse code.

notbdu: 👍

xichen2020:
Closing because this is superseded by #35 .

PR #39, opened 2018-12-11
Add bool enc/dec impl.

notbdu:
Adds bool encoder/decoder implementation. Also adds a generic template for run length encoding/decoding and tests.


encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,70 @@
...
+// Current returns the current string.
+func (rl *RunLengthValueIterator) Current() GenericValue { return rl.curr }
+
+// Err returns any error recorded while iterating.
+func (rl *RunLengthValueIterator) Err() error { return rl.err }
+
+// Close the iterator.
+func (rl *RunLengthValueIterator) Close() error {
+	rl.closed = true
+	return nil
```
xichen2020: Clear other references such as `rl.err` and `rl.reader` etc.

encoding/template/run_length_encode.go
```diff
@@ -0,0 +1,80 @@
...
+	writer io.Writer,
+	extBuf *[]byte, // extBuf is an external byte buffer for memory re-use.
+	marshalFn ValueMarshalFn,
+	valuesIt ForwardValueIterator,
+) error {
+	// Ensure that our buffer size is large enough to handle varint ops.
+	*extBuf = bytes.EnsureBufferSize(*extBuf, binary.MaxVarintLen64, bytes.DontCopyData)
+
+	var (
+		last        *GenericValue
```
xichen2020: Can you use `GenericValue` instead of `*GenericValue`? I'd think if it's a type alias of `interface{}` using `GenericValue` should work?

If you want to check for first value, you can define a separator `firstTime` var for clarity, initialize it to `true,` and set it to `false` during iteration.

encoding/template/run_length_encode.go
```diff
@@ -0,0 +1,80 @@
...
+
+		// Incrememnt repetitions and continue if we find a repetition.
+		if *last == curr {
+			repetitions++
+			continue
+		}
+
+		// last and curr don't match, write out the run length encoded repetitions
+		// and perform housekeeping.
+		n := binary.PutVarint(*extBuf, int64(repetitions))
```
xichen2020: So if a value only appears once, the `repetitions` is 0 instead of 1? Conventionally the `repetitions` should be the same as the number of times value appears so I'd think `repetitions` should rather be 1 here?

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,70 @@
...
+	reader io.Reader,
+	unmarshalFn ValueUnmarshalFn,
+) *RunLengthValueIterator {
+	return newRunLengthValueIterator(reader, unmarshalFn)
+}
+
+// RunLengthValueIterator iterates over a run length encoded stream of GenericValue data.
+type RunLengthValueIterator struct {
+	reader      io.Reader
+	unmarshalFn ValueUnmarshalFn
```
xichen2020: `readValue ReadValueFn` might be a more proper name.

encoding/template/run_length_encode.go
```diff
@@ -0,0 +1,80 @@
...
+	Next() bool
+	Err() error
+	Current() GenericValue
+}
+
+// runLengthEncodeValue run length encodes a stream of GenericValue.
+func runLengthEncodeValue(
+	writer io.Writer,
+	extBuf *[]byte, // extBuf is an external byte buffer for memory re-use.
+	marshalFn ValueMarshalFn,
```
xichen2020: `writeValue WriteValueFn` is a more proper name?

encoding/template/run_length_encode.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		*last = curr
+		repetitions = 0
+	}
+	if err := valuesIt.Err(); err != nil {
+		return err
+	}
+
+	// Encode the final value.
+	n := binary.PutVarint(*extBuf, int64(repetitions))
```
xichen2020: Hm presumably you should only encode it if the running value has not been encoded yet? This would be as easy as checking `repetitions == 0` if you use the semantics where `repetitions` means the number of times a value appears (i.e., 1 value corresponds to a `repetitions` value of 1).

encoding/bool_encode.go
```diff
@@ -1,24 +1,42 @@
...
+		t: []byte{1},
+		f: []byte{0},
+	}
+}
 
 // Encode encodes a collection of bools and writes the encoded bytes to the writer.
-func (enc *BoolEnc) Encode(writer io.Writer, values ForwardBoolIterator) error {
-	return errors.New("not implemented")
+func (enc *BoolEnc) Encode(writer io.Writer, valuesIt ForwardBoolIterator) error {
+	return runLengthEncodeBool(writer, &enc.buf, enc.marshalFn, valuesIt)
```
xichen2020: Do we want to support bit packing for booleans? Feels it could help greatly to reduce space for booleans where RLE doesn't necessarily apply.

encoding/bool_decode.go
```diff
@@ -7,3 +7,32 @@ type BoolDecoder interface {
...
+
+// BoolDec is a bool decoder.
+type BoolDec struct {
+	t byte
+}
+
+// NewBoolDecoder creates a new bool decoder.
+func NewBoolDecoder() *BoolDec {
+	return &BoolDec{
+		t: byte(1),
```
xichen2020: nit: Probably no need to cast, also can name this `trueByte` as a const.

encoding/template/run_length_decode.go
notbdu: 👍

encoding/template/run_length_encode.go
notbdu: Yea, I was doing it to catch the first value. 👍

encoding/template/run_length_encode.go
notbdu: I was thinking that it should be 0, as it's the # of times a value is repeated. And if it shows up only once, that means there are 0 repetitions.

I can change it to 1 here though.

encoding/template/run_length_decode.go
notbdu: 👍

encoding/bool_decode.go
```diff
@@ -2,8 +2,41 @@ package encoding
...
+		t: trueByte,
+	}
+}
+
+// Decode encoded bool data in a streaming fashion.
+func (dec *BoolDec) Decode(reader io.Reader) (ForwardBoolIterator, error) {
+	return runLengthDecodeBool(reader, dec.unmarshalFn), nil
+}
+
+func (dec *BoolDec) unmarshalFn(reader io.Reader) (bool, error) {
```
xichen2020: This probably should be named `readBool` for consistency?

encoding/bool_encode.go
```diff
@@ -1,24 +1,42 @@
...
+}
+
+func (enc *BoolEnc) marshalFn(writer io.Writer, value bool) error {
```
xichen2020: This probably should be named `writeBool` for consistency?

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/xichen2020/eventdb/x/io"
+)
+
+// WriteValueFn reads a GenericValue from an `io.Reader`.
+type WriteValueFn func(reader io.Reader) (GenericValue, error)
+
+// runLengthDecodeValue run length decodes a stream of GenericValues.
+func runLengthDecodeValue(
+	reader io.Reader,
+	writeValue WriteValueFn,
```
xichen2020: Shouldn't this be `readValue ReadValueFn`? 

If so, you should move `ReadValueFn` to `run_length_encode.go` and `WriteValueFn` to `run_length_decode.go`.

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,72 @@
...
+) *RunLengthValueIterator {
+	return newRunLengthValueIterator(reader, writeValue)
+}
+
+// RunLengthValueIterator iterates over a run length encoded stream of GenericValue data.
+type RunLengthValueIterator struct {
+	reader      io.Reader
+	writeValue  WriteValueFn
```
xichen2020: Should be `readValue`.

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,72 @@
...
+	err         error
+}
+
+// Next iteration.
+func (rl *RunLengthValueIterator) Next() bool {
+	if rl.closed || rl.err != nil {
+		return false
+	}
+
+	if rl.repetitions > 1 {
```
xichen2020: Hm should this be `rl.repetitions >= 1`? If so can you add a unit test for this (assuming there's one)?

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,72 @@
...
+
+	rl.curr, rl.err = rl.writeValue(rl.reader)
+	return rl.err == nil
+}
+
+// Current returns the current string.
+func (rl *RunLengthValueIterator) Current() GenericValue { return rl.curr }
+
+// Err returns any error recorded while iterating.
+func (rl *RunLengthValueIterator) Err() error { return rl.err }
```
xichen2020: Is the intention to return `io.EOF` when we exhaust the reader?

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,72 @@
...
+func (rl *RunLengthValueIterator) Close() error {
+	rl.closed = true
+	rl.err = nil
+	rl.reader = nil
+	return nil
+}
+
+func newRunLengthValueIterator(
+	reader io.Reader,
+	writeValue WriteValueFn,
```
xichen2020: `readValue ReadValueFn`.

encoding/template/run_length_encode.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/mauricelam/genny/generic"
+	"github.com/xichen2020/eventdb/x/bytes"
+)
+
+// GenericValue is a generic type.
+type GenericValue generic.Type
+
+// ReadValueFn reads a GenericValue from an `io.Reader`.
+type ReadValueFn func(writer io.Writer, value GenericValue) error
```
xichen2020: `WriteValueFn`.

encoding/template/run_length_encode.go
```diff
@@ -0,0 +1,82 @@
...
+			continue
+		}
+
+		// last and curr don't match, write out the run length encoded repetitions
+		// and perform housekeeping.
+		n := binary.PutVarint(*extBuf, int64(repetitions))
+		if _, err := writer.Write((*extBuf)[:n]); err != nil {
+			return err
+		}
+		if err := readValue(writer, last); err != nil {
```
xichen2020: `writeValue`.

encoding/template/run_length_encode.go
xichen2020: Sure, can you put these few lines into a function (e.g., `writeRLE(writer, value, repetitions)`) and reuse it here and in the for loop above?

encoding/template/run_length_decode.go
notbdu: The unit test case I have covers this. Since the logic was changed to: "if a value shows up once, the # of repetitions starts at 1." So what would happen is:
- First pass, # of repetitions is 0.
- Read in value, # of repetitions is 1.
- Read in next value, # of repetitions is 2.
- Loop once based on this condition and decrement # of repetitions.

encoding/bool_decode.go
notbdu: 👍

encoding/bool_encode.go
notbdu: 👍

encoding/template/run_length_decode.go
notbdu: 👍

encoding/template/run_length_decode.go
notbdu: 👍

encoding/template/run_length_decode.go
notbdu: I've been doing that for all readers up until now. I feel like that is the correct behaviour?

encoding/template/run_length_decode.go
notbdu: 👍

encoding/template/run_length_encode.go
notbdu: 👍

encoding/template/run_length_encode.go
notbdu: 👍

encoding/bool_encode.go
```diff
@@ -1,24 +1,42 @@
...
+	return &BoolEnc{
+		t: []byte{1},
+		f: []byte{0},
+	}
+}
 
 // Encode encodes a collection of bools and writes the encoded bytes to the writer.
-func (enc *BoolEnc) Encode(writer io.Writer, values ForwardBoolIterator) error {
-	return errors.New("not implemented")
+func (enc *BoolEnc) Encode(writer io.Writer, valuesIt ForwardBoolIterator) error {
```
xichen2020: Can you add a TODO(bodu) for implementing bit packing?

encoding/bool_encode.go
```diff
@@ -1,24 +1,42 @@
...
+	return runLengthEncodeBool(writer, &enc.buf, enc.writeBool, valuesIt)
+}
+
+func (enc *BoolEnc) writeBool(writer io.Writer, value bool) error {
+	// Write a whole byte since we won't win much by writing a single bit since
+	// varints are byte packed.
+	writeValue := enc.t
```
xichen2020: Take it or leave it, but you don't actually need to store the `t` and `f` as fields to avoid a bit of GC / allocation since you can also do sth like
```
var b [1]byte
if value {
  b[0] = trueByte
}
_, err := writer.Write(b[:])
return err
```

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,83 @@
...
+	errInvalidNumberOfRepetitions = errors.New("invalid # of repetitions < 1")
+)
+
+// ReadValueFn reads a GenericValue from an `io.Reader`.
+type ReadValueFn func(reader io.Reader) (GenericValue, error)
+
+// runLengthDecodeValue run length decodes a stream of GenericValues.
+func runLengthDecodeValue(
+	reader io.Reader,
+	ReadValue ReadValueFn,
```
xichen2020: Probably should be `readValue readValueFn`?

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,83 @@
...
+) *RunLengthValueIterator {
+	return newRunLengthValueIterator(reader, ReadValue)
+}
+
+// RunLengthValueIterator iterates over a run length encoded stream of GenericValue data.
+type RunLengthValueIterator struct {
+	reader      io.Reader
+	readValue   ReadValueFn
+	curr        GenericValue
+	repetitions int64
```
xichen2020: `int` would probably be more idiomatic.

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if rl.repetitions == 0 {
+		rl.repetitions, rl.err = binary.ReadVarint(rl.reader)
+		if rl.err != nil {
+			return false
+		}
+		rl.curr, rl.err = rl.readValue(rl.reader)
+	}
+
+	if rl.repetitions < 1 {
```
xichen2020: Should probably move this into the `if rl.repetitions == 0` block before line 46?

Basically this function can be changed to:
```
	if rl.closed || rl.err != nil {
 		return false
 	}
        if rl.repetitions > 0 {
          rl.repetitions--
          return true
        }
  	rl.repetitions, rl.err = binary.ReadVarint(rl.reader)
 	if rl.err != nil {
 	  return false
 	}
        if rl.repetitions < 1 {
            rl.err = errInvalidNumberOfRepetitions
            return false
        }
 	rl.curr, rl.err = rl.readValue(rl.reader)
         if rl.err != nil {
           return false
         }
 	rl.repetitions--
        return true
}
```

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,83 @@
...
+		rl.err = errInvalidNumberOfRepetitions
+		return false
+	}
+
+	if rl.repetitions >= 1 {
```
xichen2020: If you get here, `rl.repetitions` is guaranteed to be at least 1, so you don't need this check.

encoding/template/run_length_decode.go
xichen2020: Yeah sounds good.

encoding/template/run_length_encode.go
```diff
@@ -0,0 +1,85 @@
...
+
+		// Incrememnt repetitions and continue if we find a repetition.
+		if last == curr {
+			repetitions++
+			continue
+		}
+
+		// last and curr don't match, write out the run length encoded repetitions
+		// and perform housekeeping.
+		if err := writeRLE(writer, extBuf, writeValue, last, int64(repetitions)); err != nil {
```
xichen2020: nit: You can probably pass `*extBuf` instead of `extBuf` since you are not changing `extBuf` in `writeRLE`.

encoding/template/run_length_encode.go
```diff
@@ -0,0 +1,85 @@
...
+
+	return writeRLE(writer, extBuf, writeValue, last, int64(repetitions))
+}
+
+func writeRLE(
+	writer io.Writer,
+	extBuf *[]byte, // extBuf is an external byte buffer for memory re-use.
+	writeValue WriteValueFn,
+	value GenericValue,
+	repetitions int64,
```
xichen2020: nit: `int` is more idiomatic.

encoding/template/run_length_decode.go
notbdu: I made this `int64` so I could read directly into it using `binary.ReadVarint`.

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,83 @@
...
+	errInvalidNumberOfRepetitions = errors.New("invalid # of repetitions (< 1)")
+)
+
+// readValueFn reads a GenericValue from an `io.Reader`.
+type readValueFn func(reader io.Reader) (GenericValue, error)
+
+// runLengthDecodeValue run length decodes a stream of GenericValues.
+func runLengthDecodeValue(
+	reader io.Reader,
+	readValue readValueFn,
```
xichen2020: nit: Should probably be `readValueFn readValueFn`.

encoding/template/run_length_decode.go
xichen2020: Hm sure.

encoding/template/run_length_decode.go
```diff
@@ -0,0 +1,83 @@
...
+	rl.repetitions, rl.err = binary.ReadVarint(rl.reader)
+	if rl.err != nil {
+		return false
+	}
+	if rl.repetitions < 1 {
+		rl.err = errInvalidNumberOfRepetitions
+		return false
+	}
+
+	rl.curr, rl.err = rl.readValue(rl.reader)
```
xichen2020: You stilt want to check for `rl.err` here, i.e.
```
rl.curr, rl.err = rl.readValue(rl.reader)
         if rl.err != nil {
           return false
         }
 	rl.repetitions--
        return true
```

encoding/template/run_length_encode.go
```diff
@@ -0,0 +1,85 @@
...
+		return err
+	}
+
+	return writeRLE(writer, *extBuf, writeValue, last, repetitions)
+}
+
+func writeRLE(
+	writer io.Writer,
+	extBuf []byte, // extBuf is an external byte buffer for memory re-use.
+	writeValue writeValueFn,
```
xichen2020: `writeValueFn`

encoding/template/run_length_encode.go
```diff
@@ -0,0 +1,85 @@
...
+	Next() bool
+	Err() error
+	Current() GenericValue
+}
+
+// runLengthEncodeValue run length encodes a stream of GenericValue.
+func runLengthEncodeValue(
+	writer io.Writer,
+	extBuf *[]byte, // extBuf is an external byte buffer for memory re-use.
+	writeValue writeValueFn,
```
xichen2020: `writeValueFn`

PR #40, opened 2018-12-12
Add impl for double encoder and decoder.

notbdu:
Adds an implementation for double encoding/decoding. We are encoding the values as is using the golang stdlib binary Read/Write impl.


encoding/double_encode.go
```diff
@@ -16,9 +17,15 @@ type DoubleEncoder interface {
...
-func NewDoubleEncoder(writer io.Writer) *DoubleEnc { return &DoubleEnc{} }
+func NewDoubleEncoder() *DoubleEnc { return &DoubleEnc{} }
 
 // Encode encodes a collection of doubles and writes the encoded bytes to the writer.
-func (enc *DoubleEnc) Encode(writer io.Writer, values ForwardDoubleIterator) error {
-	return errors.New("not implemented")
+func (enc *DoubleEnc) Encode(writer io.Writer, valuesIt ForwardDoubleIterator) error {
+	// Encode doubles as 8 bytes on disk.
+	for valuesIt.Next() {
+		if err := binary.Write(writer, endianness, math.Float64bits(valuesIt.Current())); err != nil {
```
xichen2020: This incurs an allocation due to `interface{}` cast. Use `xio.WriteInt` instead.

Also I thought we got rid of `endianness`? Where do we use it?

encoding/double_decode.go
```diff
@@ -1,9 +1,65 @@
...
+	}
+}
+
+// Next iteration.
+func (it *DefaultDoubleIterator) Next() bool {
+	if it.closed || it.err != nil {
+		return false
+	}
+	var v uint64
+	it.err = binary.Read(it.reader, endianness, &v)
```
xichen2020: This incurs an allocation due to `interface{}` cast. Use `xio.ReadInt` instead.

Also I thought we got rid of `endianness`? Where do we use it?

encoding/double_decode.go
```diff
@@ -1,9 +1,65 @@
...
+}
+
+// Current returns the current double.
+func (it *DefaultDoubleIterator) Current() float64 { return it.curr }
+
+// Err returns any error recorded while iterating.
+func (it *DefaultDoubleIterator) Err() error { return it.err }
+
+// Close the iterator.
+func (it *DefaultDoubleIterator) Close() error {
```
xichen2020: `it.reader = nil` and `it.err = nil`.

encoding/double_encode.go
notbdu: I just added it to use it here. I can switch to `WriteInt`, I was trying to avoid having an internal buffer.

encoding/double_decode.go
notbdu: 👍

encoding/double_decode.go
notbdu: 👍

x/io/int.go
```diff
@@ -13,7 +13,7 @@ func WriteInt(x uint64, n int, buf []byte) {
 // Precondition: n <= 8 && len(buf) >= n
 func ReadInt(n int, buf []byte) uint64 {
 	var x uint64
-	for i := 0; i < n; i++ {
+	for i := n - 1; i >= 0; i-- {
```
xichen2020: Good catch, this is why we need unit tests lol.

PR #41, opened 2018-12-14
Add impl for time encoder/decoder and refactor.

notbdu:
Implements delta encoding/decoding for the time encoder/decoder. Refactors the delta encoding/decoding logic into a generic template for use btwn int/time encoder/decoder.


encoding/template/delta_iterator.go
```diff
@@ -44,22 +53,22 @@ func (it *DeltaIntIterator) Next() bool {
 	if isNegative {
 		// Zero out the negative bit.
 		delta &^= it.negativeBit
-		it.curr -= int(delta)
+		it.curr -= GenericValue(delta)
```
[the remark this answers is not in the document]
notbdu: I didn't know how to resolve this as the go compiler does not recognize `+/-` operations against `GenericValue`. Tried googling around to see if there was some iface I could embed into `GenericValue`. 

What would an `asValue` function do and how would that help in this situation

notbdu:
The time encoding PR is dependent on the delta encoding refactor. 

I can open one and then the other later?

encoding/template/delta_iterator.go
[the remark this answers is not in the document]
xichen2020: Yeah was thinking passing an `subFn` along the lines of
```
type subFn func(v GenericValue, delta int)
```

xichen2020:
So there's no easy way to separate the refactoring part from the time encoding/decoding logic? If so, that's fine, I'll just take a look at the larger PR then.

encoding/template/delta_iterator.go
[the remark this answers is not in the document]
notbdu: Okay, it was a little ugly but I added some math fns to take care of that and removed the build flags.

encoding/time_decode.go
```diff
@@ -1,9 +1,134 @@
...
+		// Make buf at least big enough to hold Utime64 values.
+		buf: make([]byte, uint64SizeBytes),
+		// Make a bitWriter w/ a nil buffer that will be re-used for every `Decode` call.
+		bitReader: bitstream.NewReader(nil),
+	}
+}
+
+// Decode encoded time data in a streaming fashion.
+func (dec *TimeDec) Decode(reader io.Reader) (ForwardTimeIterator, error) {
+	// Reset internal state at the beginning of every `Encode` call.
```
xichen2020: every `Decode` call.

encoding/time_decode.go
```diff
@@ -1,9 +1,134 @@
...
+		return nil, err
+	}
+
+	var (
+		iter ForwardTimeIterator
+		err  error
+	)
+	switch dec.metaProto.Encoding {
+	case encodingpb.EncodingType_DELTA:
+		iter, err = dec.decodeDelta()
```
xichen2020: Can simply `return dec.decodeDelta()` here and remove `return iter, err` below.

encoding/time_decode.go
```diff
@@ -1,9 +1,134 @@
...
+	case encodingpb.ResolutionType_SECOND:
+		resolution = time.Second
+	default:
+		return nil, fmt.Errorf("resolution (%v) is not a valid resolution", resolution)
+	}
+	return newScaledUpTimeIterator(resolution, deltaIter), nil
+}
+
+// scaledUpTimeIterator scales down the time values to the specified resolution.
+type scaledUpTimeIterator struct {
```
xichen2020: Hm not sure if `scaledUp` (or `scaledDown`) is a good name... Can we merge `scaledUpIterator` and `scaledDownIterator` into one `scaledIterator` implementation and pass a `scaleFn` to the ctor that does the scaling?

The `scaleFn` can be defined as `type scaleFn func(v int64, resolution time.Duration) int64`.

Also I wonder if it's worth moving the scaling operation from `Next` to `Current`. I imagine there may be times where we can simply skip values without actually needing to do the scaling operation, in which case doing it lazily in `Current` would help.

encoding/time_decode.go
```diff
@@ -1,9 +1,134 @@
...
+// Current returns the current int64.
+func (it *scaledUpTimeIterator) Current() int64 { return it.curr }
+
+// Err returns any error recorded while iterating.
+func (it *scaledUpTimeIterator) Err() error { return it.valuesIt.Err() }
+
+// Close the iterator.
+func (it *scaledUpTimeIterator) Close() error {
+	it.closed = true
+	valuesIt := it.valuesIt
```
xichen2020: nit: Can simplify this to
```
it.valuesIt.Close()
it.valuesIt = nil
```

encoding/time_encode.go
```diff
@@ -1,24 +1,146 @@
...
 
 import (
-	"errors"
+	"fmt"
 	"io"
+	"math"
+	"math/bits"
+	"time"
+
+	bitstream "github.com/dgryski/go-bitstream"
```
xichen2020: Import ordering.

encoding/time_encode.go
```diff
@@ -1,24 +1,146 @@
...
+	}
+}
 
 // Encode encodes a collection of time values and writes the encoded bytes to the writer.
-func (enc *TimeEnc) Encode(writer io.Writer, values ForwardTimeIterator) error {
-	return errors.New("not implemented")
+func (enc *TimeEnc) Encode(
+	writer io.Writer,
+	valuesIt RewindableTimeIterator,
+	resolution time.Duration,
```
xichen2020: Can we define a struct `EncodeTimeOptions` for this? I'd imagine we need similar constructs for block encoding too.
```
type EncodeTimeOptions struct {
  Resolution time.Duration
}

func (enc *TimeEnc) Encode(writer io.Writer, valuesIt RewindableTimeIterator, opts EncodeTimeOptions) {
  ...
}
```

encoding/time_encode.go
```diff
@@ -1,24 +1,146 @@
...
+		return fmt.Errorf("resolution (%v) is not a valid resolution", resolution)
+	}
+
+	var (
+		firstVal int64
+		max      = int64(math.MinInt64)
+		min      = int64(math.MaxInt64)
+	)
+	for idx := 0; valuesIt.Next(); idx++ {
+		curr := valuesIt.Current() / int64(resolution)
```
xichen2020: nit: move `int64(resolution)` outside the for loop.

encoding/time_encode.go
```diff
@@ -1,24 +1,146 @@
...
+	}
+	if err := valuesIt.Err(); err != nil {
+		return err
+	}
+	valuesIt.Rewind()
+
+	enc.metaProto.Encoding = encodingpb.EncodingType_DELTA
+	enc.metaProto.Resolution = resType
+	enc.metaProto.DeltaStart = firstVal
+	enc.metaProto.BitsPerEncodedValue = int64(bits.Len(uint((max - min) + 1))) // Add 1 for the sign bit.
```
xichen2020: Hm looks like it should rather be `bits.Len(uint(max - min)) + 1` instead?

encoding/time_encode.go
```diff
@@ -1,24 +1,146 @@
...
+// scaledDownTimeIterator scales down the time values to the specified resolution.
+type scaledDownTimeIterator struct {
+	resolution time.Duration
+	valuesIt   RewindableTimeIterator
+	err        error
+	curr       int64
+	closed     bool
+}
+
+func newscaledDownTimeIterator(
```
xichen2020: `newScaled...`

encoding/template/delta_iterator.go
```diff
@@ -1,34 +1,40 @@
...
 	bitstream "github.com/dgryski/go-bitstream"
 )
 
-// DeltaIntIterator iterates over a stream of delta encoded data.
-type DeltaIntIterator struct {
+// DeltaValueIterator iterates over a stream of delta encoded data.
+type DeltaValueIterator struct {
 	bitReader           *bitstream.BitReader
 	bitsPerEncodedValue int64
+	subFn               func(v GenericValue, delta int) GenericValue
```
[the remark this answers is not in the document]
notbdu: Ah, so the type signatures are the same so I'm just going to make it `applyOpToValueIntFn`.

encoding/time_encode.go
notbdu: Ah, actually I don't need another bit for the sign since we don't have negative timestamps haha. 😅

notbdu:
I'm going to rebase to resolve the conflicts.

encoding/time_encode.go
xichen2020: Hm but you can still have negative deltas no?

notbdu: Aren't timestamps incrementing only? Unless we are assuming out of order timestamps.

xichen2020: They can be out of order since these are event timestamps that may be ingested in no particular order.

notbdu: Kk, I'll update this.

encoding/template/delta_encode.go
```diff
@@ -0,0 +1,59 @@
...
+	valuesIt ForwardValueIterator,
+	subFn func(curr GenericValue, last GenericValue) int,
+) error {
+	// Encode the first value which is always a delta of 0.
+	if !valuesIt.Next() {
+		return valuesIt.Err()
+	}
+
+	// Write an extra bit to encode the sign of the delta.
+	if err := bitWriter.WriteBits(uint64(0), int(bitsPerEncodedValue)); err != nil {
```
xichen2020: Hm shouldnt this be `bitWriter.WriteBits(valuesIt.Current(), bitsPerEncodedValue)` rather?

encoding/template/run_length_decode.go
```diff
@@ -11,22 +11,22 @@ var (
 	errNonPositiveNumberOfRepetitions = errors.New("non positive number of repetitions")
 )
 
-// readValueFn reads a GenericValue from an `io.Reader`.
-type readValueFn func(reader io.Reader) (GenericValue, error)
+// readValueFn reads a GenericRLValue from an `io.Reader`.
```
xichen2020: nit: `RLValue` -> `RLEValue`.

encoding/time_decode.go
```diff
@@ -1,9 +1,88 @@
...
+		resolution = time.Millisecond
+	case encodingpb.ResolutionType_SECOND:
+		resolution = time.Second
+	default:
+		return nil, fmt.Errorf("resolution (%v) is not a valid resolution", resolution)
+	}
+	return newScaledTimeIterator(resolution, deltaIter, dec.scaleUpFn), nil
+}
+
+func (dec *TimeDec) scaleUpFn(v int64, resolution time.Duration) int64 {
```
xichen2020: The `scaleUpFn` and `scaleDownFn` should be moved into `math.go` since they are non-member functions.

generated/proto/encodingpb/time.proto
```diff
@@ -0,0 +1,23 @@
...
+    MILLISECOND = 3;
+    SECOND = 4;
+}
+
+message TimeMeta {
+    EncodingType encoding = 1;
+    ResolutionType resolution = 2;
+    // Bits per encoded value. All encoded values are bit packed.
+    int64 bits_per_encoded_value = 3;
+    int64 delta_start = 4;
```
xichen2020: Hm why is this in the proto? Shouldnit it be encoded together with the data stream?

persist/fs/writer.go
```diff
@@ -288,7 +288,8 @@ func (w *writer) writeValues(
 	case field.StringType:
 		return w.sw.Encode(writer, valueIt.stringIt)
 	case field.TimeType:
-		return w.tw.Encode(writer, valueIt.timeIt)
+		// TODO(bodu): have the resolution bubble down from the data source?
+		return w.tw.Encode(writer, valueIt.timeIt, encoding.EncodeTimeOptions{Resolution: time.Nanosecond})
```
xichen2020: This would be configured as part of storage options and passed down here, yes.

encoding/template/delta_encode.go
notbdu: The first value should be encoded as 0.

encoding/template/run_length_decode.go
notbdu: Ah, just realized that I don't need to do this. I moved all the generic types to `generic.go`.

generated/proto/encodingpb/time.proto
notbdu: This is to encode the first value so that we can store the first value as 0. 

Since we are only reserving enough bits to store the max delta, in some cases we won't have enough bits to store the full first value.

encoding/encoding.go
```diff
@@ -3,4 +3,5 @@ package encoding
 // For allocating a buffer large enough to hold uint64 values.
 const (
 	uint64SizeBytes = 8
+	uint64NumBits   = 64
```
xichen2020: nit: `uint64SizeBits = uint64SizeBytes * 8` to be consistent? Also for template code perhaps just embed the const `64` since otherwise you'll need to define this const eveywhere where the template is instantiated.

encoding/template/delta_iterator.go
```diff
@@ -0,0 +1,89 @@
...
+type DeltaValueIterator struct {
+	bitReader           *bitstream.BitReader
+	bitsPerEncodedValue int64
+	subFn               applyOpToValueIntFn
+	addFn               applyOpToValueIntFn
+	negativeBit         uint64
+	curr                GenericValue
+	err                 error
+	closed              bool
+	isFirstValue        bool
```
xichen2020: nit: the convention is to make it so that you don't need to explicitly initialize it (i.e., the default value for bool should just work by changing `isFirstValue` to `isDeltaValue`)

encoding/template/template.go
```diff
@@ -0,0 +1,3 @@
+package template
+
+const uint64NumBits = 64
```
xichen2020: Embed this.

PR #42, opened 2018-12-14
Database server read path (part 1)

xichen2020:
cc @notbdu 

This PR adds the first part of the database read path logic, primarily threading the `Query` API throughout the database package. 

Additionally I realized that on the write path I need to have a mechanism to keep all known segments or its metadata in memory, and be able to unload the segments based on certain criteria, so I added logic to the write path to keep the sealed segments sorted by their end times in a skip list so it's easy to locate the start of segments on read, as well as the `Tick` API to unload segments based on the segments' last read times and offload them if they aren't read within 5 minutes (configurable) since their last read.


storage/namespace.go
```diff
@@ -90,6 +102,40 @@ func (n *dbNamespace) Write(ev event.Event) error {
...
 }
 
+func (n *dbNamespace) Query(
+	ctx context.Context,
+	q query.ParsedQuery,
+) (query.ResultSet, error) {
+	var res query.ResultSet
+	shards := n.getOwnedShards()
+	for _, shard := range shards {
+		shardRes, err := shard.Query(ctx, q)
```
notbdu: Do we want to parallelize this or create a pool of query threads?

xichen2020: For now let's leave it as sequential, can always easily parallelize if needed. Want to measure the what the raw performance is first and can concentrate on perf optimizations afterwards.

PR #54, opened 2018-12-30
Add benchmarks for merging multiple doc id set iters

xichen2020:
cc @notbdu

This PR adds benchmarks to identify the most efficient approach for merging multiple doc ID set iterators. It also replaces `Add` with more efficient `DirectAdd`.

### DIFF
--- a/document/in_all_doc_id_set_iterator_bench_test.go
+++ b/document/in_all_doc_id_set_iterator_bench_test.go
@@ -1,0 +1,226 @@
+package document
+
+import (
+	"testing"
+
+	rroaring "github.com/RoaringBitmap/roaring"
+	proaring "github.com/pilosa/pilosa/roaring"
+	"github.com/willf/bitset"
+)
+
+func BenchmarkInAllDocIDSetIteratorDense(b *testing.B) {
+	benchmarkInAllDocIDSetIterator(b, testDenseBenchSetup)
+}
+
+func BenchmarkDocIDSetIteratorBitsetDense(b *testing.B) {
+	benchmarkDocIDSetIteratorBitset(b, testDenseBenchSetup)
+}
+
+func BenchmarkDocIDSetIteratorRRoaringBitmapDense(b *testing.B) {
+	benchmarkDocIDSetIteratorRRoaringBitmap(b, testDenseBenchSetup)
+}
+
+func BenchmarkDocIDSetIteratorPilosaRoaringBitmapDense(b *testing.B) {
+	benchmarkDocIDSetIteratorPilosaRoaringBitmap(b, testDenseBenchSetup)
+}
+
+func BenchmarkInAllDocIDSetIteratorSparse(b *testing.B) {
+	benchmarkInAllDocIDSetIterator(b, testSparseBenchSetup)
+}
+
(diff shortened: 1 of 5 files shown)


document/in_all_doc_id_set_iterator_bench_test.go
```diff
@@ -0,0 +1,214 @@
...
+			count++
+		}
+	}
+}
+
+func initBenchBitmaps(ts testBenchSetup) []*proaring.Bitmap {
+	bms := make([]*proaring.Bitmap, 0, ts.numFields)
+	for i := 0; i < ts.numFields; i++ {
+		bm := proaring.NewBitmap()
+		for j := 0; j < ts.totalDocs; j++ {
```
black-adder: nit: (totally irrelevant given this is only for testing)
```
tick := ts.everyNs[i]
for j := 0; j < ts.totalDocs; j+=tick {
   bm.DirectAdd(uint64(j))
}
```

document/in_all_doc_id_set_iterator_bench_test.go
```diff
@@ -0,0 +1,214 @@
...
+	for i := 0; i < b.N; i++ {
+		iters := make([]DocIDSetIterator, 0, ts.numFields)
+		for _, bm := range bms {
+			iters = append(iters, newbitmapBasedDocIDIter(bm.Iterator()))
+		}
+		inAllIter := newInAllDocIDSetIterIterator(iters...)
+
+		count := 0
+		for inAllIter.Next() {
+			count++
```
black-adder: why is the count needed?

document/in_all_doc_id_set_iterator_bench_test.go
```diff
@@ -0,0 +1,214 @@
...
+func benchmarkDocIDSetIteratorBitset(b *testing.B, ts testBenchSetup) {
+	bms := initBenchBitmaps(ts)
+
+	bs1 := bitset.New(uint(ts.totalDocs))
+	bs2 := bitset.New(uint(ts.totalDocs))
+	buf := make([]uint, 256)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		bs1.ClearAll()
```
black-adder: is it fair for this benchmark to reuse the same bitset and incur the extra cost of the [ClearAll()](https://github.com/willf/bitset/blob/71fa2377963fc761fc6556dd5895dff5816d4e8c/bitset.go#L327) when there are other benchmarks that are just initializing a new bitmap/iterator?

document/in_all_doc_id_set_iterator_bench_test.go
xichen2020: Was used to verify correctness, can remove.

document/in_all_doc_id_set_iterator_bench_test.go
xichen2020: I imagine the bitset will be cached inside the implementation and reused, whereas there's no API support for reusing bitmaps.

document/in_all_doc_id_set_iterator_bench_test.go
xichen2020: Fair enough, can change.

PR #64, opened 2019-01-09
Use forked skiplist impl.

notbdu:
Get the first element greater than or equal to the lookup key.

### DIFF
--- a/storage/shard.go
+++ b/storage/shard.go
@@ -13,8 +13,8 @@ import (
 	"github.com/m3db/m3x/context"
 	xerrors "github.com/m3db/m3x/errors"
 	xlog "github.com/m3db/m3x/log"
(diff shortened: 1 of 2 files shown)


storage/shard.go
```diff
@@ -159,7 +159,7 @@ func (s *dbShard) QueryRaw(
 	// TODO(xichen): Find the first element whose max time is greater than or equal
 	// to start time nanos. This is currently not implemented by the skiplist API.
 	var sealed []sealedFlushingSegment
-	geElem := s.sealedByMaxTimeAsc.Get(float64(startNanosInclusive))
+	geElem := s.sealedByMaxTimeAsc.GetGreaterThanOrEqualTo(float64(startNanosInclusive))
```
xichen2020: Please also add a test case when the value is not found.

PR #65, opened 2019-01-09
Add raw size int encode/decode option.

notbdu:
Adds a raw size encode/decode option for ints.


values/decoding/raw_size_int_iterator.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"encoding/binary"
+	"io"
+
+	xio "github.com/xichen2020/eventdb/x/io"
+)
+
+// rawSizeIntIterator iterates over a stream of
+// raw size encoded Int data.
```
xichen2020: `Int` -> `int`.

values/decoding/raw_size_int_iterator.go
```diff
@@ -0,0 +1,59 @@
...
+	if it.err != nil {
+		return false
+	}
+
+	it.curr = int(curr)
+	return true
+}
+
+// Current returns the current int.
+// NB(bodu): Caller must copy the current Int to have a valid reference between `Next()` calls.
```
xichen2020: Remove NB.

values/encoding/double_encode.go
```diff
@@ -11,6 +11,10 @@ import (
 	"github.com/xichen2020/eventdb/x/proto"
 )
 
+var (
+	uint64SizeBytes = 8
```
xichen2020: Hm is this needed?

values/encoding/int_encode.go
```diff
@@ -42,39 +41,45 @@ type intEncoder struct {
 // NewIntEncoder creates a new int encoder.
 func NewIntEncoder() IntEncoder {
 	return &intEncoder{
-		// Make buf at least big enough to hold Uint64 values.
-		buf: make([]byte, uint64SizeBytes),
+		// Make buf at least big enough to hold varint64 values.
```
xichen2020: `varint64` -> `int`.

values/encoding/int_encode.go
```diff
@@ -42,39 +41,45 @@ type intEncoder struct {
...
 	var (
-		valueMeta             = intVals.Metadata()
-		maxCardinalityAllowed = int(math.Ceil(1.0 * float64(valueMeta.Size) * intDictEncodingMaxCardinalityPercent))
-		dictionary            = make(map[int]int, maxCardinalityAllowed)
-		idx                   int
+		valueMeta                                     = intVals.Metadata()
+		maxCardinalityAllowed                         = int(math.Ceil(1.0 * float64(valueMeta.Size) * intDictEncodingMaxCardinalityPercent))
+		dictionary                                    = make(map[int]int, maxCardinalityAllowed)
+		idx                                           int
+		estimatedNumOfBytesRequiredForRawSizeEncoding int
```
xichen2020: This is a bit too verbose. Perhaps `numRawEncodingBytes`?

values/encoding/int_encode.go
```diff
@@ -89,16 +94,35 @@ func (enc *intEncoder) Encode(intVals values.IntValues, writer io.Writer) error
...
-		metaProto.BitsPerEncodedValue = int64(math.Max(float64(bits.Len(uint(len(dictionary)-1))), 1))
+		bytesPerDictionaryValue := int64(math.Ceil(float64(bits.Len(uint(valueMeta.Max-valueMeta.Min))) / 8.0))
+		bitsPerDictEncodedValue := int64(math.Max(float64(bits.Len(uint(len(dictionary)-1))), 1))
+		estimatedNumOfBytesRequiredForDictionaryEncoding := (len(dictionary) * int(bytesPerDictionaryValue)) + int((bitsPerDictEncodedValue*int64(valueMeta.Size))/8)
+
+		// Pick the best option btwn raw size and dict encoding.
+		if estimatedNumOfBytesRequiredForDictionaryEncoding < estimatedNumOfBytesRequiredForRawSizeEncoding {
+			// Min number of bytes to encode each value into the int dictionary.
+			metaProto.BytesPerDictionaryValue = bytesPerDictionaryValue
+			// Min number of bits required to encode dictionary indices. If there is only one value, use 1 bit.
```
xichen2020: Please move this comment to be above line 100.

values/encoding/int_encode.go
```diff
@@ -89,16 +94,35 @@ func (enc *intEncoder) Encode(intVals values.IntValues, writer io.Writer) error
...
+
 	if len(dictionary) <= maxCardinalityAllowed {
-		metaProto.Encoding = encodingpb.EncodingType_DICTIONARY
-		// Min number of bytes to encode each value into the int dictionary.
-		metaProto.BytesPerDictionaryValue = int64(math.Ceil(float64(bits.Len(uint(valueMeta.Max-valueMeta.Min))) / 8.0))
-		// Min number of bits required to encode dictionary indices. If there is only one value, use 1 bit.
-		metaProto.BitsPerEncodedValue = int64(math.Max(float64(bits.Len(uint(len(dictionary)-1))), 1))
+		bytesPerDictionaryValue := int64(math.Ceil(float64(bits.Len(uint(valueMeta.Max-valueMeta.Min))) / 8.0))
+		bitsPerDictEncodedValue := int64(math.Max(float64(bits.Len(uint(len(dictionary)-1))), 1))
+		estimatedNumOfBytesRequiredForDictionaryEncoding := (len(dictionary) * int(bytesPerDictionaryValue)) + int((bitsPerDictEncodedValue*int64(valueMeta.Size))/8)
```
xichen2020: `numDictionaryEncodingBytes` might be sufficient.

xichen2020: Also nit: no need for the pair of parens around the multiplication.

values/encoding/int_encode.go
```diff
@@ -89,16 +94,35 @@ func (enc *intEncoder) Encode(intVals values.IntValues, writer io.Writer) error
...
+			metaProto.Encoding = encodingpb.EncodingType_DICTIONARY
+		} else {
+			metaProto.Encoding = encodingpb.EncodingType_RAW_SIZE
+		}
 	} else {
-		metaProto.Encoding = encodingpb.EncodingType_DELTA
-		// Add 1 for the sign bit.
-		metaProto.BitsPerEncodedValue = int64(bits.Len(uint(valueMeta.Max-valueMeta.Min)) + 1)
+		bitsPerDeltaEncodedValue := int64(bits.Len(uint(valueMeta.Max-valueMeta.Min)) + 1) // Add 1 for the sign bit.
+		// 8 bytes is required for the first value.
```
xichen2020: Use `uint64SizeBytes` instead of `8`?

values/encoding/int_encode.go
```diff
@@ -89,16 +94,35 @@ func (enc *intEncoder) Encode(intVals values.IntValues, writer io.Writer) error
...
+		estimatedNumOfBytesRequiredForDeltaEncoding := 8 + int(float64(bitsPerDeltaEncodedValue*int64(valueMeta.Size-1))/8.0)
```
xichen2020: `numDeltaEncodingSize` might be sufficient.

values/encoding/int_encode.go
```diff
@@ -112,6 +136,13 @@ func (enc *intEncoder) Encode(intVals values.IntValues, writer io.Writer) error
 	defer valuesIt.Close()
 
 	switch metaProto.Encoding {
+	case encodingpb.EncodingType_RAW_SIZE:
+		if err := enc.rawSizeEncode(
```
xichen2020: Probably more readable to write this as 
```
if err := enc.rawEncode(valuesIt, writer); err != nil {
  return err
}
```

values/encoding/int_encode.go
```diff
@@ -145,6 +176,20 @@ func (enc *intEncoder) reset(writer io.Writer) {
 	enc.dictionaryProto.Data = enc.dictionaryProto.Data[:0]
 }
 
+func (enc *intEncoder) rawSizeEncode(
```
xichen2020: Probably just `rawEncode`?

values/encoding/int_encode.go
```diff
@@ -42,39 +41,45 @@ type intEncoder struct {
 // NewIntEncoder creates a new int encoder.
 func NewIntEncoder() IntEncoder {
 	return &intEncoder{
-		// Make buf at least big enough to hold Uint64 values.
-		buf: make([]byte, uint64SizeBytes),
+		// Make buf at least big enough to hold varint64 values.
+		buf: make([]byte, binary.MaxVarintLen64),
```
xichen2020: No need to store `buf` as a field, you can define `buf` on the stack in `rawEncode`:
```
var b [binary.MaxVarintLen64]byte
buf := b[:]
```
then just use `buf` where `enc.buf` is used.

x/io/int.go
```diff
@@ -53,6 +55,18 @@ func ReadVarint(data []byte) (n int64, bytesRead int, err error) {
 
 // VarintBytes returns the number of byts a varint occupies.
 func VarintBytes(v int64) int {
-	var buf [binary.MaxVarintLen64]byte
-	return binary.PutVarint(buf[:], v)
+	// Convert to unsigned integer first because that is how writes/reads
```
xichen2020: Discussed offline, either do 10 comparisons or binary search.

values/encoding/double_encode.go
notbdu: I stopped using it in another file where it was defined but it is still used in `double_encode.go` so I moved it here.

values/encoding/int_encode.go
notbdu: I'm going to change it to `varint encoded 64 bit int values`.

values/encoding/double_encode.go
xichen2020: Sounds good, can you make it a const?

values/encoding/int_encode.go
notbdu: Changed it to `varintEncode`.

values/encoding/int_encode.go
notbdu: The `buf` is used elsewhere as well. (e.g. for encoding the proto metadata).

values/encoding/int_encode.go
notbdu: Changed it to `...varint encoded 64 bit int values.`

values/encoding/double_encode.go
notbdu: I removed the usage of this constant from another file but it is still being used in `double_encode.go` so I moved the definition here.

values/decoding/raw_size_int_iterator.go
```diff
@@ -0,0 +1,59 @@
...
+	"encoding/binary"
+	"io"
+
+	xio "github.com/xichen2020/eventdb/x/io"
+)
+
+// rawSizeIntIterator iterates over a stream of
+// raw size encoded Int data.
+// TODO(xichen): Get the buffer from bytes pool.
+type rawSizeIntIterator struct {
```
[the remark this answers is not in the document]
notbdu: Went w/ `varintIntIterator`.

values/encoding/int_encode.go
xichen2020: If you use `varintEncode` above, then use `numVarintEncodingBytes` here to be consistent.

notbdu:
l ran some benchmarks against a comparison versus binary search approach and the comparison approach is more efficient. Here is the code:
```
package main

import (
	"encoding/binary"
	"math"
	"testing"
)

const (
	one   = (1 << 7) / 2
	two   = (1 << 14) / 2
	three = (1 << 21) / 2
	four  = (1 << 28) / 2
	five  = (1 << 35) / 2
	six   = (1 << 42) / 2
	seven = (1 << 49) / 2
	eight = (1 << 56) / 2
	nine  = (1 << 63) / 2
)

func calculateVarintBytesComparison(value int64) int {
	if value < one || value >= -one {
		return 1
	} else if value < two || value >= -two {
		return 2
	} else if value < three || value >= -three {
		return 3
	} else if value < four || value >= -four {
		return 4
	} else if value < five || value >= -five {
		return 5
	} else if value < six || value >= -six {
		return 6
	} else if value < seven || value >= -seven {
		return 7
	} else if value < eight || value >= -eight {
		return 8
	} else if value < nine || value >= -nine {
		return 9
	}
	return binary.MaxVarintLen64
}

type bounds struct {
	upper int
	lower int
}

var (
	lookupArray = []bounds{
		bounds{
			upper: one - 1,
			lower: -one,
		},
		bounds{
			upper: two - 1,
			lower: -two,
		},
		bounds{
			upper: three - 1,
			lower: -three,
		},
		bounds{
			upper: four - 1,
			lower: -four,
		},
		bounds{
			upper: five - 1,
			lower: -five,
		},
		bounds{
			upper: six - 1,
			lower: -six,
		},
		bounds{
			upper: seven - 1,
			lower: -seven,
		},
		bounds{
			upper: eight - 1,
			lower: -eight,
		},
		bounds{
			upper: nine - 1,
			lower: -nine,
		},
		bounds{
			upper: math.MaxInt64,
			lower: math.MinInt64,
		},
	}
)

func calculateVarintBytesBinarySearch(value int64) int {
	var (
		currIdx  int
		lowerIdx int
		upperIdx = len(lookupArray) - 1
	)
	for {
		if upperIdx == lowerIdx {
			currIdx = upperIdx
			break
		}
		// Rounds down.
		currIdx = (upperIdx + lowerIdx) / 2

		// Inside bounds.
		if value <= int64(lookupArray[currIdx].upper) && value >= int64(lookupArray[currIdx].lower) {
			upperIdx = currIdx
		} else {
			// Outside bounds.
			lowerIdx = currIdx + 1
		}
	}
	return currIdx
}

func BenchmarkComparison(b *testing.B) {
	for n := 0; n < b.N; n++ {
		calculateVarintBytesComparison(int64(n * 10))
	}
}

func BenchmarkBinarySearch(b *testing.B) {
	for n := 0; n < b.N; n++ {
		calculateVarintBytesBinarySearch(int64(n * 10))
	}
}
```

Here are the results:
```
bodu @ bodu-C02WP0EWHTDG eventdb (bdu/intrawsize) $ go test -bench=.
goos: darwin
goarch: amd64
pkg: github.com/xichen2020/eventdb
BenchmarkComparison-8           2000000000               1.38 ns/op
BenchmarkBinarySearch-8         300000000                5.25 ns/op
PASS
ok      github.com/xichen2020/eventdb   5.025s
```

xichen2020:
Sounds good, nice benckmarks, I guess the missed branch predictions and worse cache locality of binary search might have hindered its performance.

x/io/int.go
```diff
@@ -4,15 +4,26 @@ import (
...
-	"math/bits"
 )
 
 var (
 	errBufferTooSmall = errors.New("buffer too small")
 	errValueTooLarge  = errors.New("value too large")
 )
 
+// Used for calculating the number of bytes required to encode a varint.
+const (
```
xichen2020: Can we please add the benchmarks to the PR?

x/io/int.go
```diff
@@ -4,15 +4,26 @@ import (
...
+	one   = (1 << 7) / 2
```
xichen2020: Use `>>1` instead of `/ 2`

x/io/int.go
```diff
@@ -53,20 +64,26 @@ func ReadVarint(data []byte) (n int64, bytesRead int, err error) {
...
-	uv := uint64(v) << 1
-	if v < 0 {
-		uv = ^uv
-	}
-	numBits := bits.Len64(uv)
-	if numBits == 0 {
-		// Return 1 byte if the value is 0.
+	if v < one || v >= -one {
 		return 1
+	} else if v < two || v >= -two {
```
xichen2020: You don't need the `else` here on out since you are always returning early.

x/io/int.go
notbdu: Ah, I figured the compiler would optimize this away anyways and it wouldn't be calculated at runtime. So I went w/ what is easier to read.

Can change though.

x/io/int.go
```diff
@@ -11,6 +11,19 @@ var (
 	errValueTooLarge  = errors.New("value too large")
 )
 
+// Used for calculating the number of bytes required to encode a varint.
+const (
+	one   = (1 << 7) >> 1
```
xichen2020: nit: Take it or leave it, but you can make this `1 << 6` and same for the rest.

x/io/int.go
```diff
@@ -51,8 +64,34 @@ func ReadVarint(data []byte) (n int64, bytesRead int, err error) {
 	return 0, 0, errValueTooLarge
 }
 
-// VarintBytes returns the number of byts a varint occupies.
+// VarintBytes returns the number of bytes required to encode a varint.
 func VarintBytes(v int64) int {
```
xichen2020: Can we please add some tests for this and compare the results with the value returned from `PutVarint`? Something along the lines of:
```
for i := -1Bn; i < 1Bn; i++ {
  // compare the results of `VarintBytes` against `PutVarint`
}
```

notbdu: `PutVarint` is slow to test w/ a step size of 1. I'll add the tests that I ran manually where I test on the borders of the ranges to make sure those are correct.

x/io/int.go
notbdu: I like it like this for readability since varints are based 128 encoded and the reader can clearly see the 7 bits here.

values/roundtriptest/int_test.go
```diff
@@ -13,19 +13,27 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestPositiveIntDictionaryEncodeAndDecode(t *testing.T) {
+const (
+	mockDataOutlierMax = 1 << 32
+	mockDataOutlierMin = -(1 << 32)
+)
+
+func TestPositiveIntRawSizeEncodeAndDecode(t *testing.T) {
```
xichen2020: Should be `TestPositionVarintEncodeAndDecode`.

values/roundtriptest/int_test.go
```diff
@@ -13,19 +13,27 @@ import (
...
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
 	meta := values.IntValuesMetadata{
 		Size: 512,
-		Min:  1,
-		Max:  1,
+		Min:  0,
+		Max:  mockDataOutlierMax,
```
xichen2020: Is this used to determine whether we are using varint encoding?

values/roundtriptest/int_test.go
```diff
@@ -41,19 +49,22 @@ func TestPositiveIntDictionaryEncodeAndDecode(t *testing.T) {
 	testEncodeAndDecodeInt(t, data, meta, vals)
 }
 
-func TestNegativeIntDictionaryEncodeAndDecode(t *testing.T) {
+func TestNegativeIntRawSizeEncodeAndDecode(t *testing.T) {
```
xichen2020: `TestNegativeVarintEncodeAndDecode`?

values/roundtriptest/int_test.go
```diff
@@ -41,19 +49,22 @@ func TestPositiveIntDictionaryEncodeAndDecode(t *testing.T) {
...
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
 	meta := values.IntValuesMetadata{
 		Size: 512,
-		Min:  -1,
-		Max:  -1,
+		Max:  0,
+		Min:  mockDataOutlierMin,
```
xichen2020: Same question, I assume this is used to determine we should use varint encoding?

values/roundtriptest/int_test.go
```diff
@@ -69,31 +80,57 @@ func TestNegativeIntDictionaryEncodeAndDecode(t *testing.T) {
 	testEncodeAndDecodeInt(t, data, meta, vals)
 }
 
-func TestPositiveIntDeltaEncodeAndDecode(t *testing.T) {
+func TestMixedIntRawSizeEncodeAndDecode(t *testing.T) {
```
xichen2020: `TestMixedVarintEncodeAndDecode`.

x/io/int_test.go
```diff
@@ -0,0 +1,44 @@
...
+		six - 1,
+		seven - 1,
+		eight - 1,
+		nine - 1,
+	}
+	for idx := 0; idx < len(upperBounds); idx++ {
+		require.Equal(t, binary.PutVarint(buf, upperBounds[idx]), VarintBytes(upperBounds[idx]))
+		require.Equal(t, binary.PutVarint(buf, lowerBounds[idx]), VarintBytes(lowerBounds[idx]))
+		// Also step over boundaries as a sanity check.
+		require.Equal(t, binary.PutVarint(buf, upperBounds[idx]+1), VarintBytes(upperBounds[idx]+1))
```
xichen2020: Can we also test `upperBounds[idx]-1` and `lowerBounds[idx]+1`?

x/io/int_test.go
```diff
@@ -0,0 +1,44 @@
...
+		require.Equal(t, binary.PutVarint(buf, lowerBounds[idx]-1), VarintBytes(lowerBounds[idx]-1))
+	}
+	require.Equal(t, binary.PutVarint(buf, math.MaxInt64), VarintBytes(math.MaxInt64))
```
xichen2020: +1 on checking on large values.

values/roundtriptest/int_test.go
notbdu: Yes, I'll add a comment above the const defs.

values/roundtriptest/int_test.go
notbdu: Changed to `TestNegativeIntVarintEncodeAndDecode`.

values/roundtriptest/int_test.go
notbdu: Changed to `TestPositiveIntVarintEncodeAndDecode`.

values/roundtriptest/int_test.go
notbdu: Changed to `TestMixedIntVarintEncodeAndDecode`.

values/roundtriptest/int_test.go
notbdu: Added a comment above the const def.

PR #67, opened 2019-01-11
Add Fetch API to typed fields

xichen2020:
cc @notbdu  

This PR adds `Fetch` to typed field interfaces, which allows us to fetch values from typed fields for a given doc ID set determined from a doc ID set iterator passed in.


index/full_doc_id_position_iterator.go
```diff
@@ -0,0 +1,51 @@
...
+	if it.done {
+		return false
+	}
+	if !it.maskingIt.Next() {
+		it.done = true
+		return false
+	}
+	it.currPosition++
+	it.currDocID = it.maskingIt.DocID()
+	if it.currDocID < it.numTotalDocs {
```
black-adder: I might be totally misunderstanding something; I was under the assumption that doc ids are monotonically increasing, ie the DocIDSetIterator might start at 1000 and return 1000, 1001, 1002, etc. However here we're comparing against a `full doc ID set representing [0, numTotalDocs)` which can only occur naturally once given the doc ids are monotonically increasing.

index/doc_id_position_iterator.go
```diff
@@ -0,0 +1,91 @@
...
+// Next returns true if there are more items to be iterated over.
+func (it *docIDPositionIterator) Next() bool {
+	if it.backingDone || it.maskingDone {
+		return false
+	}
+	it.advanceBackingIter()
+	if it.backingDone {
+		return false
+	}
+	if it.maskingDocID == invalidDocID {
```
black-adder: can this not be done upfront in the constructor so we don't have to check it every time?

index/doc_id_set.go
```diff
@@ -15,6 +15,10 @@ type DocIDSet interface {
 	// Iter returns the document ID set iterator.
 	Iter() DocIDSetIterator
 
+	// Fetch returns the set of positions in the current doc ID set that
+	// are also in the doc ID set given by the iterator passed in.
+	Fetch(it DocIDSetIterator) DocIDPositionIterator
```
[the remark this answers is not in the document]
xichen2020: Yeah my thinking was if at some point we want to treat the doc IDs that are in the set passed in but not in the current set differently then the name `Fetch` could potentially make more sense, but I can change it to `Intersect` as well.

index/full_doc_id_position_iterator.go
xichen2020: Hm not sure I get what you meant by "can only occur naturally once"?

index/bitmap_based_doc_id_set_iterator.go
```diff
@@ -0,0 +1,38 @@
...
+
+type bitmapBasedDocIDIterator struct {
+	rit *roaring.Iterator
+
+	closed bool
+	done   bool
+	curr   int32
+}
+
+func newbitmapBasedDocIDIterator(rit *roaring.Iterator) *bitmapBasedDocIDIterator {
```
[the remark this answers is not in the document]
xichen2020: Sounds good.

index/doc_id_set.go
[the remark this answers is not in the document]
xichen2020: This is semantically Intersect though, change included in the next PR

index/full_doc_id_position_iterator.go
xichen2020: Would prefer to be a bit defensive here in case the masking iterator goes out of range.

PR #68, opened 2019-01-11
Add Iter API to field interfaces

xichen2020:
cc @black-adder 

This PR adds `Iter` API to various field interfaces so we can iterate over doc ID and value pairs in fields iteratively, needed for ordering during query execution.


index/template/field_iterator.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+}
+
+func (it *valueFieldIterator) Next() bool {
+	if it.done || it.err != nil {
+		return false
+	}
+	if !it.docIt.Next() || !it.valIt.Next() {
+		it.done = true
+		// TODO(xichen): Add doc it errors as well.
```
black-adder: this seems trivial, why punt?

xichen2020: Yeah...except that the Err() method hasn’t been added to the DocIDSetIterator interface yet. I have a TODO for that so I’ll address this as well in the same PR.

PR #69, opened 2019-01-13
Add multi field iterator and remove genclean

xichen2020:
cc @black-adder @notbdu 

This PR adds the multi-field iterator so we can iterate over multiple fields at the same time, and remove `genclean` since it has a bug that prevents the correct mock from being generated.


index/field/multi_field_iterator.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"github.com/xichen2020/eventdb/document/field"
+)
+
+// MultiFieldIterator is an iterator that iterates over multiple fields,
+// which are joined on their doc IDs. As a result, a document needs to
+// contain all fields associated with the iterator in order for it to
+// be included in the output of the multi-field iterator.
+type MultiFieldIterator struct {
```
black-adder: Can you explain the high level use case for this? One use case I see is: user wants to see all documents that must contain all paths x, y, and z so we pass in 3 iterators for path x, y, and z into this iterator and this will intersect on the doc ids.

However, what if there are multiple types for the same path? ie we pass in 4 iterators for x (int), x (double), y, and z (two iterators for path x, one for type int, one for type double). The difference here being that we don't actually need to intersect on all 4 iterators; we can first do a union of the two iterators for path x and then do an intersection with y and z. For this case, we would need both a `MultiFieldUnionIterator` and a `MultiFieldIntersectionIterator`; First you would create a `MultiFieldUnionIterator` for x and pass this iterator into `MultiFieldIntersectionIterator` with y and z.

xichen2020: Issue opened here: #70

PR #71, opened 2019-01-14
Support querying ordered/unordered raw results in immutable segments during query execution

xichen2020:
cc @black-adder @notbdu 

This PR adds support for querying ordered and unordered raw results in immutable segments during query execution.


storage/immutable_segment.go
```diff
@@ -3,6 +3,7 @@ package storage
 import (
 	"errors"
 	"fmt"
+	"sort"
```
black-adder: meta comment: this file is over 1000 lines long, should break it up

query/raw_result.go
```diff
@@ -0,0 +1,210 @@
...
+
+// RawResult is a single raw result returned from a raw query.
+// TODO(xichen): Implement `MarshalJSON` to only marshal the `Data` field without the `data tag.
+type RawResult struct {
+	Data string
+
+	// Fields for joining and sorting purposes. These fields are empty for unsorted raw results.
+	DocID         int32
+	OrderByValues []field.ValueUnion
+	OrderIdx      int
```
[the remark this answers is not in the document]
xichen2020: So this is the index of the raw result when the raw results are ordered as dictated by the query (e.g., if there are two raw results sorted by time in descending order, then the first raw result has an `OrderIdx` of 0, and the second one has an `OrderIdx` of 1). This is so that if the raw results are sorted in a different order (e.g., by doc ID), we can resort the results later on to restore the original order.

query/raw_result.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// RawResultLessThanFn compares two raw results.
+type RawResultLessThanFn func(v1, v2 RawResult) bool
+
+// NewLessThanFn creates a less than fn from a set of field value comparison functions.
+func NewLessThanFn(compareFns []field.ValueCompareFn) RawResultLessThanFn {
+	return func(v1, v2 RawResult) bool {
+		for idx, fn := range compareFns {
+			res := fn(v1.OrderByValues[idx], v2.OrderByValues[idx])
```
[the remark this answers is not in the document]
xichen2020: Yeah so this is actually a precondition that always stands, I'll add a comment.

query/raw_result.go
```diff
@@ -0,0 +1,210 @@
...
+func (h RawResultHeap) Data() []RawResult { return h.dv }
+
+// Min returns the "smallest" heap element according to the `lessThan` function.
+func (h RawResultHeap) Min() RawResult { return h.dv[0] }
+
+// Len returns the number of items in the heap.
+func (h RawResultHeap) Len() int { return len(h.dv) }
+
+// Less returns true if item `i` is less than item `j`.
+func (h RawResultHeap) Less(i, j int) bool {
```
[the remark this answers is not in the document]
xichen2020: Yeah so this is basically just implementing the same APIs in [`container/heap`](https://golang.org/pkg/container/heap/), see `Example (IntHeap)`.

index/field/doc_id_multi_field_intersect_iterator.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/xichen2020/eventdb/document/field"
+	"github.com/xichen2020/eventdb/index"
+)
+
+// DocIDMultiFieldIntersectIterator is an iterator that iterates over multiple
+// fields for values associated with documents in both the given doc ID set
+// iterator and the multi field iterator.
+type DocIDMultiFieldIntersectIterator struct {
+	docIt        index.DocIDSetIterator
+	multiFieldIt *MultiFieldIterator
```
notbdu: nit: `MultiFieldIterator` wasn't implemented in this PR but should it be called `MultiFieldIntersectIterator` instead?

index/field/doc_id_multi_field_intersect_iterator.go
```diff
@@ -0,0 +1,99 @@
...
+
+	done       bool
+	err        error
+	currDocIDs [2]int32 // 0 for docIt, 1 for multiFieldIt
+	currVals   []field.ValueUnion
+}
+
+// NewDocIDMultiFieldIntersectIterator creates a new DocIDMultiFieldIntersectIterator.
+func NewDocIDMultiFieldIntersectIterator(
+	docIt index.DocIDSetIterator,
```
notbdu: Is this `docIt` iterator the union of the Doc ID sets of the filtered values?

If it is, is the idea for the `DocIDMultiFieldIntersectIterator` to be used to produce the order/sort by fields?

query/raw_result.go
```diff
@@ -0,0 +1,215 @@
...
+			if res > 0 {
+				return true
+			}
+		}
+		return true
+	}
+}
+
+// NewRawResultHeap creates a new raw results heap.
+func NewRawResultHeap(
```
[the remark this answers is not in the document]
xichen2020: Hm this technically can be a min heap or a max heap depending on the sort ordering.

index/field/doc_id_multi_field_intersect_iterator.go
xichen2020: Sure sounds good.

index/field/doc_id_multi_field_intersect_iterator.go
xichen2020: Yes and yes.

PR #75, opened 2019-01-15
Support raw query against mutable segments

xichen2020:
cc @notbdu @black-adder 

This PR adds logic to support raw query against mutable segments.

### DIFF
--- a/storage/immutable_segment.go
+++ b/storage/immutable_segment.go
@@ -3,17 +3,13 @@ package storage
 import (
 	"errors"
 	"fmt"
-	"sort"
 	"sync"
 
 	"github.com/xichen2020/eventdb/document/field"
-	"github.com/xichen2020/eventdb/filter"
-	"github.com/xichen2020/eventdb/index"
 	indexfield "github.com/xichen2020/eventdb/index/field"
 	"github.com/xichen2020/eventdb/persist"
 	"github.com/xichen2020/eventdb/query"
 	"github.com/xichen2020/eventdb/values"
-	iterimpl "github.com/xichen2020/eventdb/values/iterator/impl"
 	"github.com/xichen2020/eventdb/x/hash"
 
 	"github.com/m3db/m3x/context"
@@ -50,6 +46,7 @@ var (
 	errFlushingNotInMemoryOnlySegment        = errors.New("flushing a segment that is not in memory only")
 	errDataNotAvailableInInMemoryOnlySegment = errors.New("data unavaible for in-memory only segment")
 	errNoTimeValuesInTimestampField          = errors.New("no time values in timestamp field")
+	errNoRawDocSourceField                   = errors.New("no raw doc source field in the segment")
 	errNoStringValuesInRawDocSourceField     = errors.New("no string values in raw doc source field")
 )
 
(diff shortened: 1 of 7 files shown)


storage/immutable_segment.go
```diff
@@ -207,6 +204,9 @@ func (s *immutableSeg) QueryRaw(
 		}
 	}()
 
+	if queryFields[1] == nil {
```
black-adder: make this a const? `rawDocQueryFieldIdx` or something

storage/mutable_segment.go
```diff
@@ -159,12 +160,87 @@ func (s *mutableSeg) Intersects(startNanosInclusive, endNanosExclusive int64) bo
...
+		for i := range queryFields {
+			if queryFields[i] != nil {
+				queryFields[i].Close()
+				queryFields[i] = nil
+			}
+		}
+	}()
+
+	numDocuments := s.mutableSegmentBase.NumDocuments()
+	s.RUnlock()
```
black-adder: just curious, if we want to give up the mutex ASAP, can we move the defer block from L188 after the mutex unlock? Will that actually make any difference?

storage/mutable_segment.go
```diff
@@ -281,6 +357,61 @@ func (s *mutableSeg) Close() {
...
+	fieldTypes []field.ValueTypeSet,
+	fieldIndexMap []int,
+	queryFields []indexfield.DocsField,
+	err error,
+) {
+	numFieldsForQuery := q.NumFieldsForQuery()
+	fieldTypes = make([]field.ValueTypeSet, numFieldsForQuery)
+	fieldIndexMap = make([]int, numFieldsForQuery)
+	queryFields = make([]indexfield.DocsField, 0, numFieldsForQuery)
+	allAllowedFieldTypes := q.AllowedFieldTypes
```
black-adder: this rename doesn't seem super necessary

storage/mutable_segment.go
```diff
@@ -281,6 +357,61 @@ func (s *mutableSeg) Close() {
...
+			// Field does not exist.
+			queryFields = append(queryFields, nil)
+			for sourceIdx := range fm.AllowedTypesBySourceIdx {
+				fieldIndexMap[sourceIdx] = len(queryFields) - 1
+			}
+			continue
+		}
+		var fieldTypeSet field.ValueTypeSet
+		if len(fm.AllowedTypesBySourceIdx) <= 1 {
+			for _, ft := range fm.AllowedTypesBySourceIdx {
```
black-adder: i like this

storage/mutable_segment.go
```diff
@@ -281,6 +357,61 @@ func (s *mutableSeg) Close() {
...
+				fieldTypeSet = ft
+				break
+			}
+		} else {
+			fieldTypeSet = make(field.ValueTypeSet, 6)
```
black-adder: const?

storage/immutable_segment.go
xichen2020: sounds good

storage/mutable_segment.go
xichen2020: Probably no difference but doesn't hurt either, can move.

storage/mutable_segment.go
xichen2020: np, can remove.

PR #76, opened 2019-01-16
Implement Add and AddBatch for raw results

xichen2020:
cc @notbdu @black-adder 

This PR implements `Add` and `AddBatch` for raw results.

### DIFF
--- a/query/raw_result.go
+++ b/query/raw_result.go
@@ -250,7 +250,7 @@ func (r *RawResults) Add(rr RawResult) {
 // For ordered raw results, the incoming batch is sorted:
 // - If the results have not reached limit yet, the incoming results are added in order
 //   until the limit is reached, after which the incoming results are inserted and
-//   the results beyong limit are dropped.
+//   the results beyond limit are dropped.
 func (r *RawResults) AddBatch(rr []RawResult) {
 	// TODO(xichen): We currently don't use `AddBatch` for unordered raw results so punt on the
 	// implementation for now. When we do, a templatized linked list might be a better choice
--- a/query/raw_result.go
+++ b/query/raw_result.go
@@ -203,7 +203,8 @@ type RawResults struct {
 	LessThanFn        RawResultLessThanFn
 	ReverseLessThanFn RawResultLessThanFn
 
-	Data []RawResult `json:"data"`
+	Data  []RawResult `json:"data"`
+	cache []RawResult
 }
 
 // IsOrdered returns true if the raw results are kept in order.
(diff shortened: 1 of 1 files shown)


query/raw_result.go
```diff
@@ -212,19 +213,98 @@ func (r *RawResults) IsOrdered() bool {
...
 // AddBatch adds a batch of raw results to the collection.
-// TODO(xichen): Implement this.
+// For unordered raw results, the incoming batch is unsorted:
+// - If the results have not reached limit yet, the incoming results are appended at the end
+//   until the limit is reached, after which the incoming results are dropped.
+// - Otherwise, the incoming results are dropped.
+// For ordered raw results, the incoming batch is sorted:
+// - If the results have not reached limit yet, the incoming results are added in order
+//   until the limit is reached, after which the incoming results are inserted and
+//   the results beyong limit are dropped.
```
black-adder: s/beyong/beyond

xichen2020: Good catch, thanks

query/raw_result.go
```diff
@@ -212,19 +213,98 @@ func (r *RawResults) IsOrdered() bool {
...
+	// implementation for now. When we do, a templatized linked list might be a better choice
+	// for the result collection.
 	panic("not implemented")
 }
 
 // AddBatch adds a batch of raw results to the collection.
-// TODO(xichen): Implement this.
+// For unordered raw results, the incoming batch is unsorted:
+// - If the results have not reached limit yet, the incoming results are appended at the end
+//   until the limit is reached, after which the incoming results are dropped.
```
notbdu: Haven't looked but should we add a `TODO` to add short circuiting logic to stop queries from executing once the result limit is reached?

query/raw_result.go
```diff
@@ -212,19 +213,98 @@ func (r *RawResults) IsOrdered() bool {
...
+		existingIdx++
+	}
+	for newIdx < len(rr) && newResultsIdx < newSize {
+		r.cache[newResultsIdx] = rr[newIdx]
+		newResultsIdx++
+		newIdx++
+	}
+
+	// Swap data array and cache array to prepare for next add.
+	tmp := r.Data
```
notbdu: `r.Data, r.cache = r.cache, r.Data` doesn't work?

query/raw_result.go
xichen2020: Yeah this has been handled in the segment query logic.

query/raw_result.go
xichen2020: Good point, will change.

PR #77, opened 2019-01-17
Grouped query types and initial implementation

xichen2020:
cc @notbdu @black-adder 

This PR defines the group query types and adds the initial logic to thread the grouped query API through the storage layer.

### DIFF
--- a/document/field/field.go
+++ b/document/field/field.go
@@ -260,6 +260,26 @@ func compareTimeNanos(v1, v2 int64) int {
 	return 0
 }
 
+// ValuesLessThanFn compares two value unions and returns true if `v1` is less than `v2`.
+type ValuesLessThanFn func(v1, v2 []ValueUnion) bool
+
+// NewValuesLessThanFn creates a less than fn from a set of field value comparison functions.
+// Precondition: len(v1) == len(compareFns) && len(v2) == len(compareFn).
+func NewValuesLessThanFn(compareFns []ValueCompareFn) ValuesLessThanFn {
+	return func(v1, v2 []ValueUnion) bool {
+		for idx, fn := range compareFns {
+			res := fn(v1[idx], v2[idx])
+			if res < 0 {
+				return false
+			}
+			if res > 0 {
+				return true
+			}
+		}
+		return true
+	}
+}
+
 // Field is an event field.
 type Field struct {
 	Path  []string
--- a/document/field/field.go
+++ b/document/field/field.go
@@ -264,20 +264,52 @@ func compareTimeNanos(v1, v2 int64) int {
 type ValuesLessThanFn func(v1, v2 []ValueUnion) bool
 
 // NewValuesLessThanFn creates a less than fn from a set of field value comparison functions.
-// Precondition: len(v1) == len(compareFns) && len(v2) == len(compareFn).
+// Precondition: len(v1) == len(compareFns) && len(v2) == len(compareFns).
 func NewValuesLessThanFn(compareFns []ValueCompareFn) ValuesLessThanFn {
 	return func(v1, v2 []ValueUnion) bool {
 		for idx, fn := range compareFns {
 			res := fn(v1[idx], v2[idx])
 			if res < 0 {
-				return false
+				return true
 			}
 			if res > 0 {
-				return true
+				return false
 			}
 		}
 		return true
 	}
+}
+
+// FilterValues return a value array excluding the given value indices.
+// Precondition: Elements in `toExcludeIndices` are unique, monotonically increasing,
+// and within range [0, len(values)).
+// Postcondition: `values` is unmodified.
+func FilterValues(values []ValueUnion, toExcludeIndices []int) []ValueUnion {
+	if len(values) == 0 || len(toExcludeIndices) == 0 {
+		return values
+	}
+	if len(values) == len(toExcludeIndices) {
+		return nil
+	}
+	var (
+		valueIdx     = 0
+		toExcludeIdx = 0
+		res          = make([]ValueUnion, 0, len(values)-len(toExcludeIndices))
+	)
+
+	for valueIdx < len(values) && toExcludeIdx < len(toExcludeIndices) {
+		if valueIdx == toExcludeIndices[toExcludeIdx] {
+			toExcludeIdx++
+			valueIdx++
+			continue
+		}
+		res = append(res, values[valueIdx])
+		valueIdx++
+	}
+	if valueIdx < len(values) {
+		res = append(res, values[valueIdx:]...)
+	}
+	return res
 }
 
 // Field is an event field.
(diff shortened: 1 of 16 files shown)


document/field/field.go
```diff
@@ -260,6 +260,26 @@ func compareTimeNanos(v1, v2 int64) int {
...
+type ValuesLessThanFn func(v1, v2 []ValueUnion) bool
+
+// NewValuesLessThanFn creates a less than fn from a set of field value comparison functions.
+// Precondition: len(v1) == len(compareFns) && len(v2) == len(compareFn).
+func NewValuesLessThanFn(compareFns []ValueCompareFn) ValuesLessThanFn {
+	return func(v1, v2 []ValueUnion) bool {
+		for idx, fn := range compareFns {
+			res := fn(v1[idx], v2[idx])
+			if res < 0 {
+				return false
```
black-adder: `res < 0` means `v1` is smaller than `v2`, should we return true here as per comment on L263?

storage/namespace.go
```diff
@@ -130,6 +136,31 @@ func (n *dbNamespace) QueryRaw(
...
+
+	res := q.NewGroupedResults()
+	shards := n.getOwnedShards()
+	for _, shard := range shards {
+		shardRes, err := shard.QueryGrouped(ctx, q)
+		if err != nil {
+			return query.GroupedResults{}, err
+		}
+		res.AddBatch(shardRes.Groups)
+		if !res.IsOrdered() && res.LimitReached() {
```
black-adder: why is the `IsOrdered` check necessary here? Isn't `LimitReached` enough signal to bail?

query/grouped_result.go
```diff
@@ -0,0 +1,99 @@
...
+type CalculatedValue struct {
+	Value       float64
+	Values      []float64
+	TransformFn TransformFn
+}
+
+// ResultGroup is a result group.
+type ResultGroup struct {
+	// Group key corresponding to the `GroupBy` fields in the query.
+	Key []field.ValueUnion
```
black-adder: Would `[]field.ValueType` more suffice here? Also, maybe rename to `GroupByKeys`?

document/field/field.go
xichen2020: Good catch, thanks! I'll add a test

storage/namespace.go
xichen2020: The idea is that if the results are ordered, even if you have gathered enough from one shard, you still want to merge w/ the results from other shards as those may be "larger" than your current results, and as such we can't bail early for ordered results. We can however do so for unordered results.

query/grouped_result.go
xichen2020: These are the keys used as the identity of each group so I don't think we can live with just a list of types? `Keys` is probably sufficient given that the struct is already named `ResultGroup`.

storage/namespace.go
black-adder: ahh, that makes sense. Mind leaving an inline comment? Perhaps even add a new function that encapsulates this check? since we use it in 3 other places.

xichen2020: Fair enough, can add a new function `IsComplete` and clarifying comments for the function, perhaps replacing `LimitReached` too.

query/grouped_result.go
black-adder: I think I'm misunderstanding something, ValueUnion looks like:
```
type ValueUnion struct {
	Type         ValueType
	BoolVal      bool
	IntVal       int
	DoubleVal    float64
	StringVal    string
	TimeNanosVal int64
}
```
not sure how we can "key" on it since there's no path info and essentially all we have is `ValueType`. The other variables seem useless for keying unless we want to GroupBy `int=40`? Is that what we're going after?

xichen2020: Yeah so the value union is to work around the limitation of lack of generics and no true union support. Imagine you have a query that groups results by a string field `foo` and a int field `bar`, then the keys for each group will be a []ValueUnion whose first element is a value union containing a string value and the second element is a value union containing an int value.

query/grouped_result.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Field values to order the groups by.
+	OrderByValues []field.ValueUnion
+
+	// A list of calculated values for a result group.
+	CalculatedValues []CalculatedValue
+}
+
+// GroupedResults is a collection of result groups.
```
notbdu: Q: Does every segment within the bounds of the query return a `ResultGroup`? And `GroupedResults` is the collection of all results from every relevant segment.

storage/shard.go
```diff
@@ -199,6 +185,79 @@ func (s *dbShard) QueryRaw(
...
+	active, sealed, cleanup := s.getEligibleSegmentsWithLock(q.StartNanosInclusive)
+	s.RUnlock()
+
+	defer cleanup()
+
+	// Querying active segment and adds to result set.
+	activeRes, err := active.QueryGrouped(ctx, q)
+	if err == errMutableSegmentAlreadySealed {
+		// The active segment has become sealed before a read can be performed
+		// against it. As a result we should retry the read.
```
notbdu: Should we wait before retrying here?

notbdu: Also, why don't we just check the accessor count and delay sealing of the active segment until the accessor count goes to 0? Or are we afraid of query contention from delaying segment sealing?

document/field/field.go
```diff
@@ -260,6 +260,26 @@ func compareTimeNanos(v1, v2 int64) int {
 	return 0
 }
 
+// ValuesLessThanFn compares two value unions and returns true if `v1` is less than `v2`.
+type ValuesLessThanFn func(v1, v2 []ValueUnion) bool
+
+// NewValuesLessThanFn creates a less than fn from a set of field value comparison functions.
+// Precondition: len(v1) == len(compareFns) && len(v2) == len(compareFn).
```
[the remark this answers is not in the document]
xichen2020: Sounds good will fix.

query/grouped_result.go
xichen2020: TBD, we can either treat them as different groups, or disallow that.

document/field/field.go
```diff
@@ -260,6 +260,26 @@ func compareTimeNanos(v1, v2 int64) int {
...
+func NewValuesLessThanFn(compareFns []ValueCompareFn) ValuesLessThanFn {
```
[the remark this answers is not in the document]
xichen2020: Hm not sure I understand this comment?

query/grouped_result.go
xichen2020: Each segment returns a set of `ResultGroups`, and the `GroupedResults` merges the results from different segments and different shards.

storage/shard.go
xichen2020: Not sure why we should wait before retrying? The error means the active segment has been changed so should be re-queried.

To prevent a race condition between checking accessor count and querying, you'll need a write lock to make sure the accessor count doesn't change while the segment is queried, no?

notbdu: Mmm that's true. Nvm then haha.

PR #78, opened 2019-01-17
Instrument Storage/Database

notbdu:
Adds instrumentation to storage/database.

### DIFF
--- a/persist/fs/manager.go
+++ b/persist/fs/manager.go
@@ -6,8 +6,6 @@ import (
 
 	"github.com/xichen2020/eventdb/index/field"
 	"github.com/xichen2020/eventdb/persist"
-
-	"github.com/m3db/m3x/clock"
 )
 
 var (
@@ -29,7 +27,6 @@ type persistManager struct {
 	opts           *Options
 	filePathPrefix string
 	writer         segmentWriter
-	nowFn          clock.NowFn
 
 	status persistManagerStatus
 	pp     persist.PreparedPersister
@@ -44,7 +41,6 @@ func NewPersistManager(opts *Options) persist.Manager {
 	pm := &persistManager{
 		opts:           opts,
 		filePathPrefix: opts.FilePathPrefix(),
-		nowFn:          opts.ClockOptions().NowFn(),
 		writer:         newSegmentWriter(opts),
 		status:         persistManagerIdle,
 	}
--- a/storage/sealed_segment.go
+++ b/storage/sealed_segment.go
@@ -55,9 +55,6 @@ type sealedSegment interface {
 
 	// Unload unloads all fields from memory.
 	Unload() error
-
-	// LoadedStatus reports whether the underlying immutable segment is loaded or not.
-	LoadedStatus() segmentLoadedStatus
 }
 
 type sealedFlushingSegment interface {
--- a/storage/sealed_segment.go
+++ b/storage/sealed_segment.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/m3db/m3x/clock"
 	"github.com/m3db/m3x/context"
+	"github.com/m3db/m3x/instrument"
 )
 
 type flushingSegment interface {
@@ -54,6 +55,9 @@ type sealedSegment interface {
 
 	// Unload unloads all fields from memory.
 	Unload() error
+
+	// LoadedStatus reports whether the underlying immutable segment is loaded or not.
+	LoadedStatus() segmentLoadedStatus
 }
 
 type sealedFlushingSegment interface {
@@ -78,6 +82,7 @@ const (
 type sealedFlushingSegmentOptions struct {
 	nowFn                clock.NowFn
 	unloadAfterUnreadFor time.Duration
+	instrumentOptions    instrument.Options
 }
 
 // Concurrent access to the sealed flushing segment is protected by the shard lock.
--- a/storage/shard.go
+++ b/storage/shard.go
@@ -67,7 +67,6 @@ type dbShardMetrics struct {
 }
 
 func newDbShardMetrics(scope tally.Scope) dbShardMetrics {
-	// TODO(bodu): possibly tag these metrics w/ the shard idx?
 	return dbShardMetrics{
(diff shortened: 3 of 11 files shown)


persist/fs/manager.go
```diff
@@ -23,6 +25,19 @@ const (
 	persistManagerPersisting
 )
 
+type persistManagerMetrics struct {
```
xichen2020: Hey I think recording latency within the persist manager might be too fined grained, it might make sense to first record the `Flush` latency at the namespace level to start with, and if we see latency issues, then we can refine the details.

Speaking of which, we should also have a gauge metric that's periodically reported that tells us when it's flushing and when it's not.

storage/immutable_segment.go
```diff
@@ -91,6 +94,29 @@ type immutableSegmentOptions struct {
 	rawDocSourcePath   []string
 	fieldHashFn        hash.StringArrayHashFn
 	fieldRetriever     persist.FieldRetriever
+	clockOptions       clock.Options
+	instrumentOptions  instrument.Options
+}
+
+type immutableSegMetrics struct {
```
xichen2020: I think this might be too fine grained for now, perhaps let's keep the metrics at the namespace level and above but not going below the shard level for now. If we see performance issues, we can always add more fine-grained metrics.

storage/mutable_segment.go
```diff
@@ -60,6 +63,25 @@ var (
 
 type isSpecialFieldFn func(fieldPath []string) bool
 
+type mutableSegMetrics struct {
```
xichen2020: Same comment as above.

storage/sealed_segment.go
```diff
@@ -68,9 +70,23 @@ const (
 	flushFailed
 )
 
+type sealedFlushingSegmentMetrics struct {
+	unload instrument.MethodMetrics
```
xichen2020: I think a counter should suffice here, i.e., it's useful to know when an unload happens, and how many times it happens, but not really how long it takes to unload (and fwiw unloading should be very fast so not quite worth recording its time).

storage/shard.go
```diff
@@ -5,6 +5,7 @@ import (
 	"math"
 	"sync"
 
+	"github.com/uber-go/tally"
```
xichen2020: Import

storage/shard.go
```diff
@@ -54,6 +55,23 @@ var (
 	errShardAlreadyClosed = errors.New("shard already closed")
 )
 
+type dbShardMetrics struct {
+	numberOfImmutableSegments tally.Gauge
+	numberOfLoadedSegments    tally.Gauge
+	numberOfUnloadedSegments  tally.Gauge
+	unload                    tally.Counter
```
xichen2020: Might want to separate this into two counters: `unloadSuccess` and `unloadErrors`

storage/shard.go
```diff
@@ -305,6 +326,25 @@ func (s *dbShard) Flush(ps persist.Persister) error {
 // a given element. Ignore for now due to lack of proper API.
 // TODO(xichen): Propagate ticking stats back up.
 func (s *dbShard) Tick(ctx context.Context) error {
+	var (
+		numberOfLoadedSegments   int
+		numberOfUnloadedSegments int
+	)
+	if err := s.forEachSegment(func(segment sealedFlushingSegment) error {
```
xichen2020: You can probably get this information returned from `tryUnloadSegments` somehow. `forEachSegment` requires locking the shard, so would be great if we can reuse the `forEachSegment` loop in `tryUnloadSegments`.

storage/shard.go
```diff
@@ -305,6 +326,25 @@ func (s *dbShard) Flush(ps persist.Persister) error {
...
+			numberOfUnloadedSegments++
+		} else {
+			numberOfLoadedSegments++
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	s.metrics.numberOfLoadedSegments.Update(float64(numberOfLoadedSegments))
```
xichen2020: There are 3 states: loaded, unloaded, and partially loaded. Can we have one counter for each?

storage/shard.go
```diff
@@ -305,6 +326,25 @@ func (s *dbShard) Flush(ps persist.Persister) error {
...
+	s.metrics.numberOfUnloadedSegments.Update(float64(numberOfUnloadedSegments))
+	s.metrics.numberOfImmutableSegments.Update(float64(s.sealedByMaxTimeAsc.Length))
```
xichen2020: This requires RLocking the shard. Is it possible to get this information from within an existing RLock in the ticking path to avoid another RLock overhead?

storage/shard.go
```diff
@@ -449,6 +494,7 @@ func (s *dbShard) tryUnloadSegments(ctx context.Context) error {
 		if err := segment.Unload(); err != nil {
 			multiErr = multiErr.Add(err)
 		}
+		s.metrics.unload.Inc(1)
```
xichen2020: Please emit counters for both successes and failures.

storage/sealed_segment.go
```diff
@@ -54,6 +55,9 @@ type sealedSegment interface {
 
 	// Unload unloads all fields from memory.
 	Unload() error
+
+	// LoadedStatus reports whether the underlying immutable segment is loaded or not.
+	LoadedStatus() segmentLoadedStatus
```
xichen2020: Remove this.

storage/shard.go
```diff
@@ -54,6 +55,27 @@ var (
...
+	numberOfImmutableSegments       tally.Gauge
+	numberOfFullyLoadedSegments     tally.Gauge
+	numberOfPartiallyLoadedSegments tally.Gauge
+	numberOfUnloadedSegments        tally.Gauge
+	unloadSuccess                   tally.Counter
+	unloadError                     tally.Counter
+}
+
+func newDbShardMetrics(scope tally.Scope) dbShardMetrics {
+	// TODO(bodu): possibly tag these metrics w/ the shard idx?
```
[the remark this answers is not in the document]
xichen2020: Remove the TODO?

PR #82, opened 2019-01-23
Use filter metadata.

notbdu:
Leverage filter metadata in queries.


values/decoding/fs_bool_values.go
```diff
@@ -38,13 +39,30 @@ func (v *fsBasedBoolValues) Iter() (iterator.ForwardBoolIterator, error) {
...
 
-// TODO(xichen): Filter implementation should take advantage of the metadata
-// to do more intelligent filtering, e.g., checking if the value is within the
-// value range.
 func (v *fsBasedBoolValues) Filter(
 	op filter.Op,
 	filterValue *field.ValueUnion,
 ) (iterator.PositionIterator, error) {
+	var (
+		numTrue  = v.Metadata().NumTrues
```
xichen2020: `numTrues`, also you can just use `v.metaProto.NumTrues` to save the cost of a function call.

values/decoding/fs_bool_values.go
```diff
@@ -38,13 +39,30 @@ func (v *fsBasedBoolValues) Iter() (iterator.ForwardBoolIterator, error) {
...
+		numFalse = v.Metadata().NumFalses
```
xichen2020: `numFalses`, and same comment as above.

values/decoding/fs_bool_values.go
```diff
@@ -38,13 +39,30 @@ func (v *fsBasedBoolValues) Iter() (iterator.ForwardBoolIterator, error) {
...
+	)
+	switch op {
```
xichen2020: Probably safer to do a nil check `filterValue` as well as checking `filterValue.Type == field.BoolType` first before the `switch`, and return an error early otherwise.

values/decoding/fs_string_values.go
```diff
@@ -60,14 +61,38 @@ func (v *fsBasedStringValues) Iter() (iterator.ForwardStringIterator, error) {
 	return newStringIteratorFromMeta(v.metaProto, v.encodedValues, v.dictArr, v.encodedDictBytes)
 }
 
-// TODO(xichen): Filter implementation should take advantage of the metadata
-// to do more intelligent filtering, e.g., checking if the value is within the
-// value range, intelligently look up filter values and bail early if not found,
-// perform filtering directly against the index to avoid string comparisons.
+// TODO(xichen): Filter implementation should intelligently look up filter values and bail early if not found
```
xichen2020: Perhaps handle the rest of the TODOs as well?

notbdu: The rest of the TODOs are more involved. Let's save them for a future PR and get these more minor changes in now.

black-adder:
@xichen2020 updated PR; that's an impressive decrease in coverage... I'll add some tests

values/decoding/fs_string_values.go
```diff
@@ -60,14 +61,35 @@ func (v *fsBasedStringValues) Iter() (iterator.ForwardStringIterator, error) {
...
+	}
+	if !op.StringIsInRange(v.metaProto, filterValue.StringVal) {
+		return impl.NewEmptyPositionIterator(), nil
+	}
+	if v.metaProto.Encoding == encodingpb.EncodingType_DICTIONARY {
+		idx, ok := v.dictMap[filterValue.StringVal]
+		if !ok {
+			return impl.NewEmptyPositionIterator(), nil
+		}
+		idxIterator, err := newIndexIteratorFromMeta(v.metaProto, v.encodedValues, v.encodedDictBytes)
```
xichen2020: Noice

values/decoding/fs_time_values.go
```diff
@@ -36,19 +38,41 @@ func (v *fsBasedTimeValues) Metadata() values.TimeValuesMetadata {
...
+	if filterValue == nil {
+		return nil, errNilFilterValue
+	}
+	if filterValue.Type != field.TimeType {
+		return nil, errUnexpectedFilterValueType
+	}
+	if !op.TimeIsInRange(v.metaProto, filterValue.TimeNanosVal) {
+		return impl.NewEmptyPositionIterator(), nil
+	}
+	// TODO(wjang): newTimeIteratorFromMeta below also calls protoResolutionToDuration. The overhead shouldn't
```
xichen2020: `TODO:(wjang)` -> `NB(wjang)` unless you plan to fix this later.

Alternatively you can have a `newTimeIteratorWithResolution` that takes in a resolution and a meta.

values/decoding/time_decode.go
```diff
@@ -45,6 +45,7 @@ func (dec *timeDecoder) DecodeRaw(data []byte) (values.CloseableTimeValues, erro
 func newTimeIteratorFromMeta(
 	metaProto encodingpb.TimeMeta,
 	encodedBytes []byte,
+	scaleUp bool,
```
xichen2020: Make this an enum perhaps? I'm a fan of more explicit parameters, e.g.,
```
type scaleMode int

const (
  noScale scaleMode = iota
  withScale
)
```
Then you can call this with `newTimeIterator(..., noScale)` which is more clear than `true` or `false`.

values/decoding/fs_time_values.go
```diff
@@ -36,19 +38,41 @@ func (v *fsBasedTimeValues) Metadata() values.TimeValuesMetadata {
...
+	// be too bad compared to the time spent iterating.
+	resolution, err := protoResolutionToDuration(v.metaProto.Resolution)
+	if err != nil {
+		return nil, err
+	}
+	timeIterator, err := newTimeIteratorFromMeta(v.metaProto, v.encodedValues, false)
+	if err != nil {
+		return nil, err
+	}
+	scaledDownFilterValue := field.NewTimeUnion(convert.ScaleDownTimeFn(filterValue.TimeNanosVal, resolution))
```
xichen2020: Add some comments here perhaps?

values/encoding/int_encode.go
```diff
@@ -96,7 +96,7 @@ func (enc *intEncoder) Encode(
 	}
 
 	if len(dictionary) <= maxCardinalityAllowed {
-		bytesPerDictionaryValue := int64(math.Ceil(float64(bits.Len(uint(valueMeta.Max-valueMeta.Min))) / 8.0))
+		bytesPerDictionaryValue := int64(math.Max(math.Ceil(float64(bits.Len(uint(valueMeta.Max-valueMeta.Min)))/8.0), 1.0))
```
black-adder: without this, if the range is 0, `bytesPerDictValue` is 0 and dictionary decoding will fail. Somehow all the dictionary encoding tests passed due to some miracle.

Also, could we encode this with bits instead of bytes?

values/decoding/int_decode.go
```diff
@@ -43,8 +44,20 @@ func (dec *intDecoder) DecodeRaw(data []byte) (values.CloseableIntValues, error)
 	if err != nil {
 		return nil, err
 	}
-
-	return newFsBasedIntValues(metaProto, data[bytesRead:], dec.dictionaryProto.Data, nRead), nil
+	var dict []int
+	if nRead > 0 {
```
xichen2020: Move this logic to `tryDecodeIntDictionary` and return `[]int` from there?

values/encoding/int_encode.go
xichen2020: Would you mind adding a few tests to make sure encoding/decoding/roundtrip an empty int/string value set works as expected? Thanks!

values/decoding/int_decode.go
```diff
@@ -65,21 +52,34 @@ func tryDecodeIntDictionary(
...
+		return 0, nil, nil
 	case encodingpb.EncodingType_DICTIONARY:
 		reader := bytes.NewReader(data)
 		bytesRead, err := xproto.DecodeIntDictionary(reader, dictionaryProto, decodeBuf)
 		if err != nil {
-			return 0, err
+			return 0, nil, err
+		}
+		var dict []int
+		if bytesRead > 0 {
```
xichen2020: nit: reduce nesting w/
```
if bytesRead == 0 {
  return 0, nil, nil
}
...
```

values/decoding/int_decode.go
```diff
@@ -112,3 +112,20 @@ func newIntIteratorFromMeta(
 		return nil, fmt.Errorf("invalid int encoding type: %v", metaProto.Encoding)
 	}
 }
+
+// newDictionaryIndexIterator creates a new int iterator that can be used to iterate over
+// dictionary index values.
+// NB: EncodingType must be encodingpb.EncodingType_DICTIONARY when calling this function.
+func newDictionaryIndexIterator(
```
xichen2020: `newIntDictionaryIndexIterator` perhaps?

values/decoding/string_decode.go
```diff
@@ -113,10 +113,10 @@ func newStringIteratorFromMeta(
 	}
 }
 
-// newIndexIteratorFromMeta creates a new int iterator from string metadata that can be used
+// newDictionaryIndexIteratorFromMeta creates a new int iterator from string metadata that can be used
 // to iterate over dictionary index values.
 // NB: EncodingType must be encodingpb.EncodingType_DICTIONARY when calling this function.
-func newIndexIteratorFromMeta(
+func newDictionaryIndexIteratorFromMeta(
```
xichen2020: `newStringDictionaryIndexIterator` perhaps?

PR #85, opened 2019-01-24
Add basic integration tests with bug fixes

black-adder:


storage/mutable_segment.go
```diff
@@ -380,6 +386,30 @@ func (s *mutableSeg) collectFieldsForRawQueryWithLock(
...
+	cleanup := func() {
+		for i := range queryFields {
+			if queryFields[i] != nil {
+				queryFields[i].Close()
+				queryFields[i] = nil
+			}
+		}
+	}
+
+	// The timestamp field and the raw doc source field must be at the head of queryFields.
```
xichen2020: Ah yeah good catch

storage/segment_query.go
```diff
@@ -324,9 +324,10 @@ func collectTopNRawResultDocIDOrderByValues(
 	// Sort the result heap in place, and when done the items are sorted from left to
 	// in the right order based on the query sorting criteria (i.e., if the sort order
 	// is ascending, the leftmost item is the smallest item).
-	for results.Len() > 0 {
-		results.Pop()
-	}
+	// TODO(wjang): This just pops every item off the heap and returns an empty slice.
```
xichen2020: As discussed offline, this should be done in a `Sort` API of the raw results heap. You can reassign this to me.

storage/shard.go
```diff
@@ -166,6 +166,10 @@ func (s *dbShard) QueryRaw(
 	if err != nil {
 		return nil, err
 	}
+	if res == nil {
```
xichen2020: Thanks fo the catch. I feel fhis should be done inside the `QueryRaw` method of the mutable segment, i.e., maintaining the invariant that if there is no error, the result is always non nil.

values/iterator/impl/array_based_bool_iterator.go
```diff
@@ -20,7 +20,7 @@ func (it *ArrayBasedBoolIterator) Next() bool {
 		return false
 	}
 	it.idx++
-	return it.idx >= len(it.values)
+	return it.idx < len(it.values)
```
xichen2020: lol good catch.

integration/config/config_01.yaml
```diff
@@ -0,0 +1,116 @@
+http:
+  listenAddress: 0.0.0.0:5678
```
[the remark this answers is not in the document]
black-adder: this is a bit of a cop out but doing this was more involved than I expected, I added a TODO; going to punt for now since this is a nice to have.

black-adder:
looking into CI failures

integration/setup.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+const (
+	gracefulShutdownTimeout = 15 * time.Second
+)
+
+var (
+	logger = log.NullLogger
+
+	// basePath contains the path to this directory.
```
xichen2020: Hm I don't like the hack.

Can you just assign the yaml file to a static string var and unmarshal from there? Something like
```
var testConfig1 =
...
```
and unmarshal directly from there?

I don't think it's important to test the file loading logic in the integration test.

integration/setup.go
```diff
@@ -0,0 +1,161 @@
...
+	db   storage.Database
+	opts instrument.Options
+	cfg  configuration
+
+	// Signals.
+	doneCh   chan struct{}
+	closedCh chan struct{}
+}
+
+func newTestServerSetup(t *testing.T, configFile string) *testServerSetup {
```
xichen2020: You can pass in the test config string here instead of the file path. All the file loading logic does is just reading the file and do
```
unmarshal := yaml.UnmarshalStrict
		if opts.DisableUnmarshalStrict {
			unmarshal = yaml.Unmarshal
		}
		if err := unmarshal(data, config); err != nil {
			return err
		}
return validator.Validate(config)
```
which you can probably create a function `loadConfig(str string)` to do so.

integration/setup.go
```diff
@@ -0,0 +1,161 @@
...
+		t.Fatalf("error loading config file %s: %v\n", configFile, err)
+	}
+
+	iOpts := instrument.NewOptions().
+		SetMetricsScope(tally.NoopScope).
+		SetLogger(logger)
+
+	namespaces, err := cfg.Database.NewNamespacesMetadata()
+	if err != nil {
+		t.Fatalf("error creating namespaces metadata: %v", err)
```
xichen2020: nit: Just a FYI we usually use `require` in `testify` for assertions etc.

integration/setup.go
```diff
@@ -0,0 +1,161 @@
...
+	// Signals.
+	doneCh   chan struct{}
+	closedCh chan struct{}
+}
+
+func newTestServerSetup(t *testing.T, configFile string) *testServerSetup {
+	configFile = basePath + configFile
+	var cfg configuration
+	if err := xconfig.LoadFile(&cfg, configFile, xconfig.Options{}); err != nil {
+		t.Fatalf("error loading config file %s: %v\n", configFile, err)
```
xichen2020: Why not return an error and have the caller assert no error?

integration/setup.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (ts *testServerSetup) stopServer(t *testing.T) {
+	close(ts.doneCh)
+
+	select {
+	case <-ts.closedCh:
+		t.Log("server closed clean")
+	case <-time.After(gracefulShutdownTimeout):
+		t.Logf("server closed due to %s timeout", gracefulShutdownTimeout.String())
```
xichen2020: nit: Use `%v` so you don't need to do `.String`.

integration/setup.go
```diff
@@ -0,0 +1,161 @@
...
+
+func (ts *testServerSetup) close(t *testing.T) {
+	ts.stopServer(t)
+	ts.stopDB(t)
+}
+
+func (ts *testServerSetup) newClient() client {
+	client := newClient(ts.cfg.HTTP.ListenAddress)
+	for {
+		if client.serverIsHealthy() {
```
xichen2020: This should be done in the test instead doing it in the client ctor.

Probably better to do the following for better separation of concerns.
```
func (ts *testServerSetup) newClient() client {
  return newClient(ts.cfg.HTTP.ListenAddress)
}

func (ts *testServerSetup) waitUntil(timeout time.Duration, condition func() bool) error {
  start := ts.nowFn()
  for !condition {
    time.Sleep(time.Millisecond * 10)
    if dur := ts.nowFn().Sub(start); dur >= timeout {
      return errTimeoutWaitingForCondition
  }
  return nil
}
```

integration/setup.go
```diff
@@ -0,0 +1,161 @@
...
+			break
+		}
+		time.Sleep(time.Millisecond * 10)
+	}
+	return client
+}
+
+func (ts *testServerSetup) writeTestFixture(t *testing.T) {
```
xichen2020: Hm this reads the input config and writes it to fixture? Probably just store the config in a var as suggested above as this seems really hacky.

integration/setup.go
```diff
@@ -0,0 +1,161 @@
...
+	data, err := ioutil.ReadFile(input)
+	if err != nil {
+		t.Fatalf("cannot read %s: %v", input, err)
+	}
+	if err := client.write(data); err != nil {
+		t.Fatalf("failed write to server: %v", err)
+	}
+}
+
+func init() {
```
xichen2020: See comments above.

query/grouped_result.go
```diff
@@ -6,6 +6,7 @@ import (
 )
 
 // GroupedResults is a collection of result groups.
+// TODO(wjang): add JSON marshaling / unmarshaling
```
xichen2020: Comments start w/ upper case and should end w/ a period as they are complete sentences.

query/raw_result.go
```diff
@@ -192,6 +192,7 @@ func (h RawResultHeap) heapify(i, n int) {
 }
 
 // RawResults is a collection of raw results.
+// TODO(wjang): add JSON marshaling / unmarshaling
```
xichen2020: Comments start w/ upper case and should end w/ a period as they are complete sentences.

storage/immutable_segment.go
```diff
@@ -221,6 +221,7 @@ func (s *immutableSeg) QueryRaw(
 	if queryFields[rawDocSourceFieldIdx] == nil {
 		return errNoRawDocSourceField
 	}
+	// TODO(wjang): this is not guaranteed, see fix in mutable_segment.go
```
xichen2020: Remove the TODO since you already fixed it in `mutable_segment.go`?

values/roundtriptest/doc.go
```diff
@@ -0,0 +1,3 @@
+package roundtriptest
+
+// Without this file, this directory fails with 'no non-test Go files'
```
xichen2020: End comments w/ a period.

What failed with `no non-test Go files`?

storage/mutable_segment.go
```diff
@@ -380,6 +386,30 @@ func (s *mutableSeg) collectFieldsForRawQueryWithLock(
...
+	fm := q.FieldConstraints[s.timestampFieldHash]
```
xichen2020: Actually I think I have a cleaner fix in a followup PR #89 , perhaps we can get that merged first and you can rebase?

xichen2020: Basically there's no requirement on maintaining this ordering, users should use `fieldIndexMap` to map the source index to the query field index, and use the mapped index to index into `queryFields`.

Makefile
```diff
@@ -70,7 +70,7 @@ define TOOL_RULES
 .PHONY: $(TOOL)
 $(TOOL): setup
 	@echo Building $(TOOL)
-	$(VENDOR_ENV) go build -o $(BUILD)/$(TOOL) ./tools/$(TOOL)/.
+	$(VENDOR_ENV) go build -o $(BUILD)/$(TOOL) ./tools/$(TOOL)/main/.
```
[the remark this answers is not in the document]
xichen2020: Hm ok weird, in that case can you move `tools/benchdb/main.go` to `tools/benchdb/main/main.go`?

values/roundtriptest/doc.go
xichen2020: kk np sounds good then.

Makefile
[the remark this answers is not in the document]
black-adder: that's already been done.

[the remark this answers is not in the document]
xichen2020: Ah sorry missed it, thanks

storage/mutable_segment.go
```diff
@@ -186,18 +186,21 @@ func (s *mutableSeg) QueryRaw(
 		return nil, errMutableSegmentAlreadySealed
 	}
 
+	numDocuments := s.mutableSegmentBase.NumDocuments()
+	if numDocuments == 0 {
+		s.RUnlock()
+		return q.NewRawResults(), nil
```
black-adder: without this, code continues to execute until `applyFilters` which will fail and throw: `no time values in timestamp field` (no docs => no timestamp index)

storage/namespace.go
```diff
@@ -129,12 +129,21 @@ func (n *dbNamespace) QueryRaw(
 		if err != nil {
 			return nil, err
 		}
-		res.MergeInPlace(shardRes)
+		if res == nil {
+			res = shardRes
+		} else {
+			if err := res.MergeInPlace(shardRes); err != nil {
```
black-adder: `q.NewRawResults().MergeInPlace(shardRes)` results in `merging two raw results with different order by field types` error since the other has an order field while the former doesn't

integration/config.go
```diff
@@ -0,0 +1,159 @@
...
+      - count: 1600
+        capacity: 262144
+      - count: 1600
+        capacity: 524288
+    watermark:
+      low: 0.7
+      high: 1.0
+`
+
+	testData1 = `{"service":"testNamespace","@timestamp":"2019-01-22T13:25:42-08:00","st":true,"sid":1,"tt":"active","tz":-6,"v":1.5}
```
xichen2020: nit: Can you align this with the rest?

integration/integration_test.go
```diff
@@ -0,0 +1,64 @@
...
+					]
+				}
+			`,
+			expectedResults: 10,
+		},
+	}
+	ts := newTestServerSetup(t, testConfig1, testData1)
+	ts.startServer()
+	defer ts.close(t)
+	ts.writeTestFixture(t)
```
xichen2020: Do you still need this?

integration/setup.go
```diff
@@ -0,0 +1,130 @@
...
+
+var (
+	logger = log.NullLogger
+)
+
+type testServerSetup struct {
+	db       storage.Database
+	opts     instrument.Options
+	cfg      configuration
+	testData string
```
xichen2020: You don't need to store the `testData` here.

integration/setup.go
xichen2020: Remove this perhaps? There is no need to stash the test data inside the test setup. You can just call `client.write([]byte(testData))` inside the test so this is redundant.

integration/setup.go
```diff
@@ -0,0 +1,130 @@
...
+	client := ts.newClient()
+	require.NoError(t, ts.waitUntil(10*time.Second, client.serverIsHealthy))
+	require.NoError(t, client.write([]byte(ts.testData)))
+}
+
+func loadConfig(t *testing.T, config string) configuration {
+	var cfg configuration
+	err := yaml.UnmarshalStrict([]byte(config), &cfg)
+	require.NoError(t, err)
+	return cfg
```
xichen2020: You need to validate this too using yaml validator by calling `validator.Validate(..)`

storage/mutable_segment.go
xichen2020: Yeah fair enough.

query/raw_result.go
```diff
@@ -247,6 +251,9 @@ func (r *RawResults) MergeInPlace(other *RawResults) error {
 	if other == nil {
 		return nil
 	}
+	if other.IsEmpty() {
+		return nil
+	}
```
xichen2020: Also 
```
if r.IsEmpty() {
  *r = *other
  other.Clear()
  return nil
}
```

storage/namespace.go
```diff
@@ -120,7 +120,7 @@ func (n *dbNamespace) QueryRaw(
 		q.StartNanosInclusive = retentionStartNanos
 	}
 
-	res := q.NewRawResults()
+	var res *query.RawResults
```
xichen2020: If you do what's suggested in `raw_result.go` you don't need to change these.

values/roundtriptest/doc.go
xichen2020: Ping on the period.

query/raw_result.go
```diff
@@ -247,6 +251,14 @@ func (r *RawResults) MergeInPlace(other *RawResults) error {
 	if other == nil {
```
xichen2020: nit: `if other == nil || other.IsEmpty()`

PR #86, opened 2019-01-24
Fix heap sorting

xichen2020:
cc @black-adder @notbdu 

This PR fixes the logic to sort the raw result heap.

### DIFF
--- a/query/raw_result.go
+++ b/query/raw_result.go
@@ -146,8 +146,9 @@ func (h *RawResultHeap) Pop() RawResult {
 	return val
 }
 
-// SortInPlace sorts the heap in place and returns the sorted data,
-// with the smallest element at the end of the returned array.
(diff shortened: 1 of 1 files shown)


query/raw_result.go
```diff
@@ -149,6 +146,20 @@ func (h *RawResultHeap) Pop() RawResult {
 	return val
 }
 
+// SortInPlace sorts the heap in place and returns the sorted data,
+// with the smallest element at the end of the returned array.
+// NB: The heap becomes invalid after this is called.
+func (h *RawResultHeap) SortInPlace() []RawResult {
+	numElems := len(h.dv)
+	for len(h.dv) > 0 {
+		h.Pop()
```
notbdu: Could use a comment here.

xichen2020: Will do.

PR #89, opened 2019-01-28
Support groupBy query

xichen2020:
cc @notbdu @black-adder 

This PR adds logic to support groupBy query in immutable segments and mutable segments.


query/raw_result.go
```diff
@@ -341,10 +243,25 @@ func (r *RawResults) AddBatch(rr []RawResult) {
...
 // Precondition: The current raw results and the other raw results are generated from the same query.
-// TODO(xichen): Validate the `OrderByFieldTypes` are the same in two result set for consistent ordering.
-func (r *RawResults) MergeInPlace(other *RawResults) {
+func (r *RawResults) MergeInPlace(other *RawResults) error {
 	if other == nil {
-		return
+		return nil
+	}
+	if !r.OrderByFieldTypes.Equal(other.OrderByFieldTypes) {
+		return fmt.Errorf("merging two raw rsults with different order by field types %v and %v", r.OrderByFieldTypes, other.OrderByFieldTypes)
```
black-adder: typo

storage/segment_query.go
```diff
@@ -654,3 +596,44 @@ func collectUnorderedMultiFieldGroupByResults(
...
 	return groupByCalcIter.Err()
 }
+
+func newSingleTypeFieldIterator(
+	fieldPath []string,
+	allowedTypes field.ValueTypeSet,
+	queryField indexfield.DocsField,
+	expectedType *field.ValueType,
+) (indexfield.BaseFieldIterator, field.ValueType, error) {
+	if len(allowedTypes) != 1 {
```
black-adder: why not just pass into `field.ValueType` instead?

calculation/result.go
```diff
@@ -315,6 +315,14 @@ func (arr ResultArray) New() ResultArray {
 	return resArray
 }
 
+// MergeInPlace merges the other result array into the current array in place.
+// Precondition: len(arr) == len(other).
+func (arr ResultArray) MergeInPlace(other ResultArray) {
```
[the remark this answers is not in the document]
xichen2020: Sounds good.

query/raw_result.go
xichen2020: ta

storage/segment_query.go
xichen2020: Wouldn't that require we do the logic from line 606 to line 613 in every place before calling `newSingleTypeFieldIterator`? This also validates that if there are multiple types for this field in the segment that error is caught here.

PR #91, opened 2019-01-28
Fix shard close panic

xichen2020:
cc @black-adder @notbdu 

This PR fixes the panic on shard close and adds a test to prevent regression.

### DIFF
--- a/storage/shard.go
+++ b/storage/shard.go
@@ -320,8 +320,8 @@ func (s *dbShard) Close() error {
 	s.unflushed = nil
 	active := s.active
 	s.active = nil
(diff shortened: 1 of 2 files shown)


storage/shard.go
```diff
@@ -320,8 +320,8 @@ func (s *dbShard) Close() error {
 	s.unflushed = nil
 	active := s.active
 	s.active = nil
-	s.sealedByMaxTimeAsc = nil
 	byMaxTimeAsc := s.sealedByMaxTimeAsc
+	s.sealedByMaxTimeAsc = nil
```
black-adder: hee hee

PR #94, opened 2019-01-30
Support time bucket query

xichen2020:
cc @black-adder @notbdu 

This PR adds support for time bucket query.


query/query.go
```diff
@@ -844,6 +914,115 @@ func (q *ParsedGroupedQuery) computeNewCalculationResultArrayFn() calculation.Ne
 	}
 }
```
black-adder: meta, this file is 1000+ lines, could break it up into 4 separate files? one for parsing, and one of each for raw, group, and time bucket?

query/query.go
```diff
@@ -844,6 +914,115 @@ func (q *ParsedGroupedQuery) computeNewCalculationResultArrayFn() calculation.Ne
...
+		Granularity:         *q.TimeGranularity,
+		Filters:             q.Filters,
+	}
+	if err := tbq.computeDerived(q.opts); err != nil {
+		return ParsedTimeBucketQuery{}, err
+	}
+	return tbq, nil
+}
+
+// NewTimeBucketResults creats a new time bucket results from the parsed time bucket query.
```
black-adder: creats

query/query.go
```diff
@@ -844,6 +914,115 @@ func (q *ParsedGroupedQuery) computeNewCalculationResultArrayFn() calculation.Ne
...
+func (q *ParsedTimeBucketQuery) NewTimeBucketResults() *TimeBucketResults {
+	bucketSizeNanos := q.Granularity.Nanoseconds()
+	startInclusiveBucketNanos := q.StartNanosInclusive / bucketSizeNanos * bucketSizeNanos
```
black-adder: doesn't 1/bucketSizeNanos * bucketSizeNanos cancel out?

black-adder: oh, you're removing the trailing nanos... a comment would be nice

storage/segment_query.go
```diff
@@ -562,6 +562,31 @@ func collectMultiFieldGroupByResults(
...
+	res *query.TimeBucketResults,
+) error {
+	timestampIter, err := timestampField.Fetch(maskingDocIDSetIter)
+	if err != nil {
+		maskingDocIDSetIter.Close()
+		return fmt.Errorf("error fetching timestamp data: %v", err)
+	}
+	defer timestampIter.Close()
+
+	// TODO(xichen): This decodes the timestamp field again if the timestamp field is loaded from
```
black-adder: i'm not following this comment, i'll ask you offline.

query/time_bucket_result.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/xichen2020/eventdb/document/field"
+)
+
+// TimeBucketResults is a collection of time buckets recording the counts
+// for documents falling in each bucket.
+type TimeBucketResults struct {
+	StartBucketNanos int64 // Inclusive
+	BucketSizeNanos  int64
+	NumBuckets       int
```
notbdu: nit: we should be able to just use `len(buckets)` instead of having another attr.

query/time_bucket_result.go
```diff
@@ -0,0 +1,102 @@
...
+		other.Clear()
+		return nil
+	}
+	if r.StartBucketNanos != other.StartBucketNanos {
+		return fmt.Errorf("merging two time bucket results with different start bucket nanos %d and %d", r.StartBucketNanos, other.StartBucketNanos)
+	}
+	if r.BucketSizeNanos != other.BucketSizeNanos {
+		return fmt.Errorf("merging two time bucket results with different bucket sizes %d and %d", r.BucketSizeNanos, other.BucketSizeNanos)
+	}
+	if r.NumBuckets != other.NumBuckets {
```
notbdu: Same comment as above regarding using `len(buckets)` instead.

storage/segment_query.go
notbdu: We're decoding once to apply filters to get the masking DocID iterator and again here when we collect the results.

query/time_bucket_result.go
notbdu: Also don't see where `NumBuckets` gets assigned a value.

xichen2020:
Re @black-adder : yes it's the equivalent of a `date historgram` in ES

query/query.go
xichen2020: Sounds good.

query/query.go
xichen2020: Sure can add comments

query/time_bucket_result.go
xichen2020: So It's assigned in `query.go`, the reason why we have `NumBuckets` is so that the buckets are allocated lazily.

query/time_bucket_result.go
xichen2020: See above.

PR #95, opened 2019-01-31
Add multi field union iterator

xichen2020:
cc @black-adder @notbdu 

This PR adds support for multi-field union iterator, and refactors the multi-field intersect iterator to reuse the `inAnyDocIDSetIterator` and `inAllDocIDSetIterator` to reduce repeated logic.

### DIFF
--- a/index/field/multi_field_union_iterator.go
+++ b/index/field/multi_field_union_iterator.go
@@ -44,7 +44,7 @@ func (it *MultiFieldUnionIterator) DocID() int32 { return it.unionIter.DocID() }
 // value array is valid.
 func (it *MultiFieldUnionIterator) Values() ([]field.ValueUnion, []bool) {
 	for i := 0; i < len(it.iters); i++ {
-		if it.unionIter.IsDoneAt(i) || it.unionIter.DocID() != it.iters[i].DocID() {
+		if !it.unionIter.IsValidAt(i) {
 			it.hasVals[i] = false
 		} else {
 			it.hasVals[i] = true
--- a/index/in_any_doc_id_set_iterator.go
+++ b/index/in_any_doc_id_set_iterator.go
@@ -87,9 +87,10 @@ func (it *InAnyDocIDSetIterator) DocID() int32 {
 	return it.docIDs[it.minDocIDIdx]
 }
 
-// IsDoneAt returns whether the iterator at a given index is done.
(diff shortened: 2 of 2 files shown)


index/field/multi_field_intersect_iterator.go
```diff
@@ -2,91 +2,56 @@ package field
...
-		}
-		if !it.iters[minIdx].Next() {
-			it.done = true
-			it.err = it.iters[minIdx].Err()
-			return false
-		}
-		it.currDocIDs[minIdx] = it.iters[minIdx].DocID()
-	}
-}
+func (it *MultiFieldIntersectIterator) Next() bool { return it.intersectIter.Next() }
```
notbdu: Looking at the implementation of `InAnyDocIDSetIterator`, it looks like it does not perform the intersect operation. As in, it only advances each underlying base iterator forward w/o ensuring that all DocIDs line up.

It looks like it's used below in the `MultiFieldUnionIterator` which makes sense. But in that case, aren't we missing the intersect operation here? Maybe I'm missing something though.

xichen2020: Note that this uses `InAllDocIDSetIterator` which advances all doc ID iterators in sync.

notbdu: Oh woops disregard my comment then.

notbdu:
LGTM.

index/field/multi_field_union_iterator.go
```diff
@@ -0,0 +1,67 @@
...
+func (it *MultiFieldUnionIterator) DocID() int32 { return it.unionIter.DocID() }
+
+// Values returns the current list of field values, which remains valid until the
+// next iteration. If the caller needs to retain a valid refence to the value array
+// after `Next` is called again, the caller needs to make a copy of the value array.
+// Caller should check the boolean array to see whether the corresponding slot in the
+// value array is valid.
+func (it *MultiFieldUnionIterator) Values() ([]field.ValueUnion, []bool) {
+	for i := 0; i < len(it.iters); i++ {
+		if it.unionIter.IsDoneAt(i) || it.unionIter.DocID() != it.iters[i].DocID() {
```
black-adder: can this whole check be encapsulated inside `InAnyDocIDSetIterator`? `it.unionIter.IsValidAt` or something?

xichen2020: Fair enough, will move

PR #98, opened 2019-02-01
Support calculation against a subset of fields during groupBy queries

### LINKED ISSUE #93: Group by query should not require all calculation fields be present
Currently if a segment does not contain all the calculation fields, it is not processed for the group by query, but we should still process it as long as it contains any of the calculation fields.

xichen2020:
cc @black-adder @notbdu 

This PR adds logic to perform calculation within a group when only a subset of the fields are available for that group (e.g., grouping by `foo` calculating `sum` for `bar` and `avg` for `baz`, and only `bar` is available for some groups, and only `baz` is available for some other groups).

Fixes #93 .


document/field/value.go
```diff
@@ -390,3 +390,57 @@ func (v Values) Clone() Values {
 	}
 	return cloned
 }
+
+// OptionalType is a type "option" which is either "null" or has a valid type.
+// It is similar to a `*ValueType` from a functionality perspective but has
+// lower GC overhead.
+type OptionalType struct {
```
notbdu: Would it be better to just have a `null` `ValueType` instead of this `HasType` `bool`? Also, if we keep the `bool`, should we rename it to `HasValidType`

storage/segment_query.go
```diff
@@ -349,7 +352,7 @@ func collectGroupedResults(
 	q query.ParsedGroupedQuery,
 	res *query.GroupedResults,
 ) error {
-	filteredGroupByCalcIterator, err := createFilteredGroupByCalcIterator(
+	filteredDocIDIter, groupByIter, calcIter, err := createFilteredGroupByCalcIterator(
```
notbdu: What's the rationale for splitting these up?

storage/segment_query.go
```diff
@@ -484,39 +505,42 @@ func collectSingleFieldGroupByResults(
...
 		)
 		for i, calc := range res.Calculations {
 			if !calc.Op.RequiresField() {
 				calcResults[i].Add(emptyValueUnion)
-			} else {
-				cv := toCalcValueFns[calcFieldIdx](&calcVals[calcFieldIdx])
-				calcResults[i].Add(cv)
-				calcFieldIdx++
+			} else if !hasVals[calcFieldIdx] {
+				// The calculation field does not have value for this iteration.
```
black-adder: you're going to have to explain this one to me offline, it looks like we want to do something like:
```
for i, calc := range res.Calculations {
    for calcFieldIdx, calcFieldIdx < len(calcVals); calcFieldIdx++ {
       // run calc for every value that exists
    }
}
```

document/field/value.go
xichen2020: `NullType` is an actual type (i.e., fields with null values) so don't think we should mix this use case with that. Thought about using `UnknownType` but just feels this is more explicitly.

`HasType` feels more succinct so going to keep it unless you feel strongly.

storage/segment_query.go
xichen2020: See the `collectGroupBy...` functions, the `filteredDocIDIter` does not have APIs to expose the underlying values but rather performs the function of moving the `groupByIter` and the `calcIter` in lock step, whereas the individual `groupByIter` and `calcIter` provide APIs to underlying field values.

document/field/value.go
```diff
@@ -390,3 +390,57 @@ func (v Values) Clone() Values {
...
+		return nil
+	}
+	if !t.HasType {
+		*t = other
+		return nil
+	}
+	if t.Type != other.Type {
+		return fmt.Errorf("merging two incompatible optional types %v and %v", *t, other)
+	}
+	return nil
```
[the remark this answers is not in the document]
xichen2020: When you get here the two types are exactly the same so there's no need to merge.

storage/immutable_segment.go
```diff
@@ -191,14 +191,33 @@ func (s *immutableSeg) QueryRaw(
...
+	// Check if this segment has all orderBy fields as they are required, and bail early
+	// otherwise.
+	for _, ob := range q.OrderBy {
+		fieldHash := s.fieldHashFn(ob.FieldPath)
+		if _, exists := s.entries[fieldHash]; !exists {
+			return nil
+		}
+	}
+
+	if res.HasOrderedFilter() && res.Len() > 0 && res.LimitReached() {
```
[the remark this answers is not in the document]
xichen2020: I guess, although you can have a limit of 0, so just wanna be extra safe here.

PR #101, opened 2019-02-06
Add a duration type that can be unmarshalled from JSON

xichen2020:
cc @black-adder @notbdu 

This PR adds a `xtime.Duration` type to facilitate unmarshalling time durations from JSON. Also fixes a panic when only `time_range` is specified in query and adds a comprehensive unit test.


xichen2020:
Apparently this is an oversight in Go 1 and will be added in Go 2.

PR #104, opened 2019-02-07
Add benchmark for reading a segment from disk

black-adder:
Q: should I just generate the 10MB file once and upload it into the repo so the benchmarks are consistent?

Technically, given the "random" isn't random, this is producing the same segment every test...


xichen2020:
Yeah that should be good enough.

black-adder:
It ran just once, so I removed the useless `b.N` loop.

PR #105, opened 2019-02-07
Add benchmark for zstd compression/decompression

notbdu:
Uses the same streaming API and zstd go lib that we use in the underlying encoder/decoders.


benchmarks/benchmarks.go
```diff
@@ -0,0 +1 @@
+package benchmarks
```
xichen2020: Move the benchmarks under `values/encoding`?

benchmarks/compression_bench_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	writer := gozstd.NewWriter(out)
+	writer.Write(data)
+	return nil
+}
+
+func decompressData(data []byte) error {
+	reader := gozstd.NewReader(bytes.NewReader(data))
+	buf := make([]byte, decompressBufferSize)
+	var (
```
xichen2020: nit: `var err error`.

benchmarks/compression_bench_test.go
```diff
@@ -0,0 +1,69 @@
...
+		err error
+	)
+	for err != io.EOF {
+		_, err = reader.Read(buf)
+	}
+	return nil
```
xichen2020: `return err` maybe?

benchmarks/compression_bench_test.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func BenchmarkCompression(b *testing.B) {
+	data, err := ioutil.ReadFile(testDataFilePath)
+	if err != nil {
+		log.Fatal(err)
```
xichen2020: Curious, can you do `require.NoError(b, ...)` etc?

benchmarks/compression_bench_test.go
```diff
@@ -0,0 +1,69 @@
...
+		b.Fatalf("failed to read in test data file (%s): %v", testDataFilePath, err)
```
xichen2020: You probably don't need `log.Fatal` if you have this?

benchmarks/compression_bench_test.go
```diff
@@ -0,0 +1,69 @@
...
+		if err := compressData(data); err != nil {
+			b.Fatalf("failed to compress to data file: %v", err)
+		}
+	}
+}
+
+func BenchmarkDecompression(b *testing.B) {
+	data, err := ioutil.ReadFile(compressedDataFilePath)
+	if err != nil {
+		log.Fatal(err)
```
xichen2020: Use `b.Fatalf("failed to read file: %v", err)`?

benchmarks/compression_bench_test.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	for n := 0; n < b.N; n++ {
+		if err := decompressData(data); err != nil {
+			log.Fatal(err)
```
xichen2020: Use `b.Fatalf("failed to decompress data: %v", err)`?

notbdu:
`require.NoError(b, ...)` works so I'll use that instead of `b.Fatalf(...)`.

values/encoding/compression_bench_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/valyala/gozstd"
+)
+
+const (
+	testDataFilePath       = "./testdata/testdata.json"
+	compressedDataFilePath = "../decoding/testdata/testdata.json.compressed"
```
xichen2020: Probably use a temporary file path using `https://golang.org/pkg/io/ioutil/#TempFile`?

values/encoding/compression_bench_test.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func compressData(data []byte) error {
+	out, err := os.OpenFile(compressedDataFilePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
```
xichen2020: Should delete the temp file created with `defer os.Remove...`

values/encoding/compression_bench_test.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	writer := gozstd.NewWriter(out)
+	_, err = writer.Write(data)
+	defer writer.Close()
```
xichen2020: No need for defer and should check close error and return if not nil.

values/encoding/compression_bench_test.go
```diff
@@ -10,22 +10,23 @@ import (
...
 
 const (
-	testDataFilePath       = "./testdata/testdata.json"
-	compressedDataFilePath = "../decoding/testdata/testdata.json.compressed"
+	testDataFilePath = "./testdata/testdata.json"
 )
 
 func compressData(data []byte) error {
-	out, err := os.OpenFile(compressedDataFilePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+	out, err := ioutil.TempFile("/tmp", "*")
```
xichen2020: nit: Just pass in an empty string as the first arg since `/tmp` may not exist on some machines, the pattern perhaps is better to use `*.decompressed`?

PR #111, opened 2019-02-07
Perform housekeeping on zstd compressed reader/writers.

notbdu:
Performs housekeeping ops to release resources used by gozstd's compressed reader/writer.

### DIFF
--- a/values/decoding/compressed_reader.go
+++ b/values/decoding/compressed_reader.go
@@ -30,3 +30,10 @@ func (cr *compressedReader) ReadByte() (byte, error) {
 	}
 	return cr.buf[0], nil
 }
+
+// Close releases the underlying resources of the compressed reader.
+func (cr *compressedReader) Close() error {
+	// Release does not return an error.
+	cr.Release()
+	return nil
+}
--- a/values/decoding/compressed_reader.go
+++ b/values/decoding/compressed_reader.go
@@ -1,9 +1,14 @@
 package decoding
 
 import (
+	"errors"
 	"io"
 
 	"github.com/valyala/gozstd"
+)
+
+var (
+	errCompressedReaderClosed = errors.New("compressedReader is closed")
 )
 
 // compressedReader reads compressed data.
@@ -14,6 +19,8 @@ type compressedReader struct {
 	// we can read in a single byte at a time to
 	// impl the `io.ByteReader` iface.
 	buf [1]byte
+
+	closed bool
 }
 
 // newCompressedReader returns a new CompressReader instance.
@@ -25,6 +32,9 @@ func newCompressedReader(reader io.Reader) *compressedReader {
 
 // ReadByte read a byte.
 func (cr *compressedReader) ReadByte() (byte, error) {
+	if cr.closed {
+		return 0, errCompressedReaderClosed
+	}
 	if _, err := cr.Read(cr.buf[:]); err != nil {
 		return 0, err
 	}
@@ -33,6 +43,11 @@ func (cr *compressedReader) ReadByte() (byte, error) {
 
 // Close releases the underlying resources of the compressed reader.
 func (cr *compressedReader) Close() {
+	if cr.closed {
+		return
+	}
+
 	// Release does not return an error.
 	cr.Release()
+	cr.closed = true
 }
--- a/values/decoding/compressed_reader.go
+++ b/values/decoding/compressed_reader.go
@@ -32,8 +32,7 @@ func (cr *compressedReader) ReadByte() (byte, error) {
 }
 
 // Close releases the underlying resources of the compressed reader.
-func (cr *compressedReader) Close() error {
+func (cr *compressedReader) Close() {
 	// Release does not return an error.
 	cr.Release()
-	return nil
 }
--- a/values/decoding/dictionary_based_string_iterator.go
+++ b/values/decoding/dictionary_based_string_iterator.go
@@ -11,8 +11,8 @@ import (
 // dictionaryBasedStringIterator iterates over a
 // dict encoded stream of string data.
 type dictionaryBasedStringIterator struct {
-	reader     xio.SimpleReadCloser
-	byteReader io.ByteReader // Same as `reader` but has the proper type to save interface conversions in `Next`
+	reader     xio.SimpleReadCloser // `reader`is used to `Close` the underlying reader and allow for resource clean up.
+	byteReader io.ByteReader        // Same as `reader` but has the proper type to save interface conversions in `Next`
 	// extDict is passed externally from the string decoder
 	// and should not be mutated during iteration.
 	extDict []string
--- a/values/decoding/dictionary_based_string_iterator.go
+++ b/values/decoding/dictionary_based_string_iterator.go
@@ -11,7 +11,8 @@ import (
 // dictionaryBasedStringIterator iterates over a
 // dict encoded stream of string data.
 type dictionaryBasedStringIterator struct {
-	reader xio.SimpleReadCloser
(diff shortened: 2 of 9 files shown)


values/decoding/compressed_reader.go
```diff
@@ -30,3 +30,10 @@ func (cr *compressedReader) ReadByte() (byte, error) {
 	}
 	return cr.buf[0], nil
 }
+
+// Close releases the underlying resources of the compressed reader.
+func (cr *compressedReader) Close() error {
```
xichen2020: Some protection to make sure we don't double close/release?

values/decoding/dictionary_based_string_iterator.go
```diff
@@ -65,5 +65,13 @@ func (it *dictionaryBasedStringIterator) Err() error {
 func (it *dictionaryBasedStringIterator) Close() {
 	it.extDict = nil
 	it.err = nil
+	// Close the underlying reader if it satisifies the `io.ReadCloser` iface.
+	rc, ok := it.reader.(io.ReadCloser)
```
xichen2020: On a second thought, this is a bit worrisome compared to the approach where we implement a wrapper `SimpleReadCloser` that has a no op `Close` as this requires we understand where we need to cast and where otherwise.

Feels it's cleaner to define in `x/io/reader.go` 
```
import (
  "io"

  xio "github.com/m3db/m3x/io"
)

type SimpleReadCloser interface {
  io.Reader
  xio.SimpleCloser
}

type ReaderNoopCloser io.Reader

// Close is a no-op satistifying the `xio.SimpleCloser` interface.
func (r ReaderNoopCloser) Close() {}
```

Then implement this interface in `compressedReader` as well.

Finally pass `xio.SimpleReadCloser` to the raw/dictionary string iterator ctors (and where else closing reader is expected), casting `io.Reader` to `xio.ReaderNoopCloser` as needed.

values/encoding/string_encode.go
```diff
@@ -112,12 +111,26 @@ func (enc *stringEncoder) Encode(strVals values.StringValues, writer io.Writer)
...
-		return enc.dictionaryEncode(valuesIt, dictionary, writer)
+		if err := enc.dictionaryEncode(valuesIt, dictionary, writer); err != nil {
+			return err
+		}
 	default:
 		return fmt.Errorf("invalid encoding type: %v", metaProto.Encoding)
 	}
+
+	// Close the writer if it satisifies the `io.WriteCloser` iface.
+	wc, ok := writer.(io.WriteCloser)
```
xichen2020: It's probably more efficient if you pull `compressWriter` declaration outside the `switch` block and do a nil check here than doing a type cast.

values/decoding/raw_size_string_iterator.go
```diff
@@ -17,21 +17,19 @@ const (
 // raw size encoded string data.
 // TODO(xichen): Get the buffer from bytes pool.
 type rawSizeStringIterator struct {
-	reader     xio.Reader
-	byteReader io.ByteReader // Same as `reader` but has the proper type to save interface conversions in `Next`
+	reader xio.SimpleReadCloser
```
xichen2020: Please keep the `byteReader`, this is to save the interface conversion cost at line 43.

values/decoding/varint_int_iterator.go
```diff
@@ -3,19 +3,21 @@ package decoding
...
 	"io"
+
+	xio "github.com/xichen2020/eventdb/x/io"
 )
 
 // varintIntIterator iterates over a stream of
 // varint encoded int data.
 type varintIntIterator struct {
-	reader io.ByteReader
+	reader xio.SimpleReadCloser
```
xichen2020: Store a separate byteReader to save the interface conversion cost at line 31. See values/decoding/raw_size_string_iterator.go

values/encoding/string_encode.go
```diff
@@ -112,12 +111,25 @@ func (enc *stringEncoder) Encode(strVals values.StringValues, writer io.Writer)
...
-		return enc.dictionaryEncode(valuesIt, dictionary, writer)
+		if err := enc.dictionaryEncode(valuesIt, dictionary, writer); err != nil {
+			return err
+		}
 	default:
 		return fmt.Errorf("invalid encoding type: %v", metaProto.Encoding)
 	}
+
+	// Close the compressWriter if its present.
+	if compressWriter != nil {
```
xichen2020: Any reason not to `defer compressWriter.Close` at line 100?

values/encoding/string_encode.go
```diff
@@ -112,12 +111,25 @@ func (enc *stringEncoder) Encode(strVals values.StringValues, writer io.Writer)
 
 	switch metaProto.Encoding {
 	case encodingpb.EncodingType_RAW_SIZE:
-		return enc.rawSizeEncode(valuesIt, writer)
+		if err := enc.rawSizeEncode(valuesIt, writer); err != nil {
```
xichen2020: Why not just `return enc.rawSizeEncode(valuesIt, writer)`?

values/encoding/string_encode.go
notbdu: Because we wanted to surface the error that it returns to the caller.

notbdu: Because we wanted to surface the error that the writer returns to the caller.

values/encoding/string_encode.go
notbdu: Same note as above, so we can surface the error from calling `compressWriter.Close`.

values/decoding/raw_size_string_iterator.go
notbdu: Where is the interface conversion? We aren't doing any further casting. Also, we're not doing this in any other location just here.

Both `io.ByteReader` and `xio.SimpleReadCloser` ifaces all implement the `ReadByte() (byte, error)` method.

values/decoding/varint_int_iterator.go
notbdu: Same comment as above.

values/decoding/raw_size_string_iterator.go
notbdu: Spoke offline, will-readd. Downcasting happens in the `ReadVarint` fn call.

values/encoding/string_encode.go
```diff
@@ -110,14 +110,24 @@ func (enc *stringEncoder) Encode(strVals values.StringValues, writer io.Writer)
 	}
 	defer valuesIt.Close()
 
+	var multiErr xerrors.MultiError
 	switch metaProto.Encoding {
 	case encodingpb.EncodingType_RAW_SIZE:
-		return enc.rawSizeEncode(valuesIt, writer)
+		multiErr.Add(enc.rawSizeEncode(valuesIt, writer))
```
xichen2020: Remember to do `multiErr = multiErr.Add` in all three places.

values/encoding/string_encode.go
```diff
@@ -110,14 +110,24 @@ func (enc *stringEncoder) Encode(strVals values.StringValues, writer io.Writer)
...
 	default:
-		return fmt.Errorf("invalid encoding type: %v", metaProto.Encoding)
+		multiErr.Add(fmt.Errorf("invalid encoding type: %v", metaProto.Encoding))
 	}
+
+	// Close the compressWriter if its present.
+	if compressWriter != nil {
+		// NB(xichen): Close flushes and closes the compressed writer but doesn't
+		// close the writer wrapped by the compressed writer.
+		multiErr.Add(compressWriter.Close())
```
xichen2020: Same here.

Issue #108, opened 2019-02-07
Data unavailable for in-memory segment error


notbdu:
Seeing this error in the logingester query logs: 
```
21:01:44.548013[W] error executing query: received HTTP status: 500 Internal Server Error. body: {"state":"Error","error":"error performing grouped query {rtapi 1549486904543668347 1549573304543668347 [] [[@timestamp]] [{[] COUNT}] [] 10 0x874db0 map[12889413958949747480:{[@timestamp] map[1:map[double:{} string:{} time:{} null:{} bool:{} int:{}] 0:map[time:{}]]}]} against database namespace rtapi: data unavaible for in-memory only segment"
```

Judging from the comment below, this should not happen. Dig into this.
```
if dataLocation == inMemoryOnly {
        // We have fields that are in the in-memory metadata hash, and the actual data
        // is not in memory, and yet the location indicates all data are in memory. This
        // is a logical error and should never happen.
        cleanup()
        return nil, errDataNotAvailableInInMemoryOnlySegment
}
```

notbdu:
After some digging, this is related to #107.

notbdu:
Attempting to write a reproducible functional test case for both.

notbdu:
Have working functional test case that repros the problem. Attempting to fix.

PR #113, opened 2019-02-08
Fix flushing logic and add functional test case.

### LINKED ISSUE #107: DB should not flush to closed segments
```
17:23:09.331474[I] creating database...
17:23:13.502315[I] starting HTTP server...
17:23:13.510196[I] http server: listening on 0.0.0.0:5678
18:35:40.229348[E] error when flushing data: namespace rtapi failed to flush data: shard 1 failed to flush data: segment writer is closed
shard 2 failed to flush data: segment writer is closed
shard 3 failed to flush data: segment writer is closed
shard 4 failed to flush data: segment writer is closed
shard 5 failed to flush data: segment writer is closed
shard 6 failed to flush data: segment writer is closed
shard 7 failed to flush data: segment writer is closed
18:36:40.229586[E] error when flushing data: namespace rtapi failed to flush data: shard 1 failed to flush data: segment writer is closed
shard 2 failed to flush data: segment writer is closed
shard 3 failed to flush data: segment writer is closed
shard 4 failed to flush data: segment writer is closed
shard 5 failed to flush data: segment writer is closed
shard 6 failed to flush data: segment writer is closed
shard 7 failed to flush data: segment writer is closed
18:37:40.229907[E] error when flushing data: namespace rtapi failed to flush data: shard 1 failed to flush data: segment writer is closed
shard 2 failed to flush data: segment writer is closed
shard 3 failed to flush data: segment writer is closed
shard 4 failed to flush data: segment writer is closed
shard 5 failed to flush data: segment writer is closed
shard 6 failed to flush data: segment writer is closed
shard 7 failed to flush data: segment writer is closed
```

notbdu:
Fixes #107 and #108. 

Also adds a functional test case that repros both issues.


persist/fs/writer.go
```diff
@@ -27,18 +27,22 @@ var (
 
 // segmentWriter is responsible for writing segments to filesystem.
 type segmentWriter interface {
-	// Open opens the writer.
-	Open(opts writerOpenOptions) error
+	// Start persisting data to disk.
```
xichen2020: nit: Should be a sentence, e.g., `Start starts persisting a segment`.

persist/fs/writer.go
```diff
@@ -27,18 +27,22 @@ var (
...
+	Start(opts writerStartOptions) error
+
+	// Finish persisting data to disk.
```
xichen2020: `Finish finishes persisting a segment and performs cleanups as necessary`.

storage/database.go
```diff
@@ -250,6 +250,13 @@ func (d *db) Close() error {
 	d.state = databaseClosed
 
 	// Close database-level resources.
+	if err := d.mediator.Close(); err != nil {
+		return err
+	}
+
+	if err := d.Options().PersistManager().Close(); err != nil {
```
xichen2020: This may need to wait till the current ticking finishes (which may perform a flush). Is that the case at the moment?

storage/flush_test.go
```diff
@@ -0,0 +1,26 @@
+package storage
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFlush(t *testing.T) {
```
xichen2020: Fine to leave this here for now, but this looks like an integration test rather than a unit test (which typically tests local functionalities than performing an end-to-end test).

storage/database.go
xichen2020: On a second thought this probably should just be removed, as closing the manager will cause the it to panic if it's in the middle of a flush, and all closing does is just setting some nil pointers (which doesn't matter much if the database is shutting down).

storage/flush_test.go
notbdu: I put it here to access internal structs to force a flush.

storage/database.go
notbdu: Removing the close completely.

PR #114, opened 2019-02-08
Implement MarshalJSON for query results

### LINKED ISSUE #112: Query unsupported type calculation.NewResultArrayFromValueTypesFn
Pulled this from the logingester logs:
```
02:07:09.546747[I] executing query [{type aggregation} {body {
  "namespace": "rtapi",
  "calculations": [
    {
      "op": "COUNT"
    }
  ],
  "time_range": "24h",
  "time_unit": 1,
  "group_by": [
    "@timestamp"
  ]
}}]
02:07:09.549066[W] error executing query: received HTTP status: 500 Internal Server Error. body: {"state":"Error","error":"json: unsupported type: calculation.NewResultArrayFromValueTypesFn"}
```

xichen2020:
cc @notbdu @black-adder 

This PR implements `MarshalJSON` for the various types of query results.

Fixes #112


calculation/value.go
```diff
@@ -16,18 +18,40 @@ const (
 	StringType
 )
 
+var (
+	nullBytes = []byte("null")
```
notbdu: Do we want to define this somewhere in a shared location?

query/result_group.go
```diff
@@ -5,15 +5,28 @@ import (
 	"github.com/xichen2020/eventdb/x/compare"
 )
 
+type nullResultGruop struct {
```
black-adder: typo

query/result_group.go
```diff
@@ -5,15 +5,28 @@ import (
...
+	Key    *string                 `json:"key"` // This should always be a nil pointer to represent null key
+	Values calculation.ResultArray `json:"values"`
+}
+
+type nullResultGruopsJSON struct {
```
black-adder: ditto

query/single_key_result_gruops_test.go
```diff
@@ -0,0 +1,114 @@
+package query
```
black-adder: file name typo

query/time_bucket_result.go
```diff
@@ -60,3 +61,30 @@ func (r *TimeBucketResults) MergeInPlace(other *TimeBucketResults) error {
...
+	}
+	res := timeBucketResultsJSON{
+		Granularity: r.BucketSizeNanos,
+		Buckets:     buckets,
+	}
+	return json.Marshal(res)
+}
+
+type timeBucketJSON struct {
+	StartAt int64 `json:"startAt"` // In nanoseconds
```
black-adder: startAtNanos?

calculation/value.go
xichen2020: Hm a bit overkill given this is only used in two places in two separate packages.

query/time_bucket_result.go
xichen2020: Sure.

PR #121, opened 2019-02-11
Require calculations w/ a group by clause.

### LINKED ISSUE #120: Group by queries should require calculations
Currently, there is no explicit requirement to supply calculations when supplying a group by condition. The implementation requires calculations w/ group by queries. Until we support returning a set of fields w/ a group by clause, this will probably be the case (outside the scope of the POC).

notbdu:
Fixes #120 .


query/unparsed_query.go
```diff
@@ -318,6 +319,9 @@ func (q *UnparsedQuery) validateGroupBy() error {
 	if q.TimeGranularity != nil {
 		return errTimeGranularityWithGroupBy
 	}
+	if len(q.Calculations) == 0 {
```
xichen2020: Perhaps move this to `parseCalculations`? Tend to keep this validation logic more manageable by only validating fields above not below.

PR #127, opened 2019-02-13
Add groupby query integration test.

### LINKED ISSUE #151: Filters on group by queries returning nil results
Both single key and multi key group queries returning nil results if a filter is specified.

```
--- FAIL: TestGroupByQueryMultiKeyNoFilterNoOrder (1.10s)
    groupby_query_multi_key_no_filter_no_order_test.go:132:
                Error Trace:    groupby_query_multi_key_no_filter_no_order_test.go:132
                Error:          Not equal:
                                expected: integration.groupByQueryMultiKeyResults{Groups:[]integration.multiKeyResultGroup{integration.multiKeyResultGroup{Key:[]interface {}{false, "inactive"}, Values:[]interface {}{10}}, integration.multiKeyResultGroup{Key:[]interface {}{true, "active"}, Values:[]interface {}{10}}}}
                                actual  : integration.groupByQueryMultiKeyResults{Groups:[]integration.multiKeyResultGroup(nil)}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,22 +1,3 @@
                                 (integration.groupByQueryMultiKeyResults) {
                                - Groups: ([]integration.multiKeyResultGroup) (len=2) {
                                -  (integration.multiKeyResultGroup) {
                                -   Key: ([]interface {}) (len=2) {
                                -    (bool) false,
                                -    (string) (len=8) "inactive"
                                -   },
                                -   Values: ([]interface {}) (len=1) {
                                -    (float64) 10
                                -   }
                                -  },
                                -  (integration.multiKeyResultGroup) {
                                -   Key: ([]interface {}) (len=2) {
                                -    (bool) true,
                                -    (string) (len=6) "active"
                                -   },
                                -   Values: ([]interface {}) (len=1) {
                                -    (float64) 10
                                -   }
                                -  }
                                - }
                                + Groups: ([]integration.multiKeyResultGroup) <nil>
                                 }
                Test:           TestGroupByQueryMultiKeyNoFilterNoOrder
FAIL
FAIL    github.com/xichen2020/eventdb/integration       1.131s
```

notbdu:
Adds a single integration test for a groupby query. 

Also fixes #151.


integration/client.go
```diff
@@ -69,6 +69,23 @@ func (c client) queryRaw(data []byte) (rawQueryResults, error) {
 	return results, nil
 }
 
+func (c client) queryGroupBy(data []byte) (groupByQueryResults, error) {
```
xichen2020: Can you pull out the common parts and move it to a separate `doQuery` function and reuse it in `queryRaw` and `queryGrouped`?

integration/groupby_query_test.go
```diff
@@ -0,0 +1,85 @@
+// +build integration
```
xichen2020: Rename this file to `groupby_query_no_filter_with_calculations_no_order.go`?

integration/groupby_query_test.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGroupByQuery(t *testing.T) {
```
xichen2020: `TestGroupByQueryNoFilterWithCalculationsNoOrder`?

integration/groupby_query_test.go
notbdu: I wanted to put many different types of group by queries in the same test file so I made the test file name generic.

integration/groupby_query_test.go
notbdu: See comment above on adding diff types of queries in the tests.

integration/groupby_query_test.go
notbdu: Spoke w/ xi offline, this is the standard pattern for writing integration tests.

notbdu: Changing the name to `groupby_query_no_filter_no_order.go` since calculations are required when you specify a group by condition so adding that to the name does not add any more information.

notbdu:
Going to change it to use gRPC only and not HTTP.

PR #129, opened 2019-02-19
Add integration tests for time bucket queries

xichen2020:
cc @notbdu @black-adder @cw9 

This PR adds integration tests for time bucket queries.

### DIFF
--- a/query/raw_query.go
+++ b/query/raw_query.go
@@ -39,6 +39,15 @@ func newParsedRawQuery(q *ParsedQuery) (ParsedRawQuery, error) {
 	return rq, nil
 }
 
+// TimestampFieldIndex returns the index of the timestamp field.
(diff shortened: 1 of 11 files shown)


query/raw_query.go
```diff
@@ -39,6 +39,15 @@ func newParsedRawQuery(q *ParsedQuery) (ParsedRawQuery, error) {
 	return rq, nil
 }
 
+// TimestampFieldIndex returns the index of the timestamp field.
+func (q *ParsedRawQuery) TimestampFieldIndex() int { return 0 }
```
cw9: nit: Should we comment these magic numbers?

xichen2020: This is actually documented in the `computeFieldConstraints` function for the respective queries.

PR #131, opened 2019-02-20
Clear string pointers before returning them to the memory pool.

notbdu:
String pointers are now cleared before putting string arrays back into the memory pool.

### DIFF
--- a/index/field/docs_field.go
+++ b/index/field/docs_field.go
@@ -855,7 +855,7 @@ func (b *docsFieldBuilder) addDouble(docID int32, v float64) error {
 func (b *docsFieldBuilder) addString(docID int32, v string) error {
 	if b.sfb == nil {
 		docIDsBuilder := b.newDocIDSetBuilder()
-		stringValuesBuilder := impl.NewArrayBasedStringValues(b.opts.StringArrayPool(), b.opts.StringValuesResetFn())
+		stringValuesBuilder := impl.NewArrayBasedStringValues(b.opts.StringArrayPool(), b.opts.StringArrayResetFn())
 		b.sfb = newStringFieldBuilder(docIDsBuilder, stringValuesBuilder)
 	}
 	return b.sfb.Add(docID, v)
@@ -872,12 +872,12 @@ func (b *docsFieldBuilder) addTime(docID int32, v int64) error {
 
 // DocsFieldBuilderOptions provide a set of options for the field builder.
 type DocsFieldBuilderOptions struct {
-	boolArrayPool       *pool.BucketizedBoolArrayPool
-	intArrayPool        *pool.BucketizedIntArrayPool
-	doubleArrayPool     *pool.BucketizedFloat64ArrayPool
-	stringArrayPool     *pool.BucketizedStringArrayPool
-	int64ArrayPool      *pool.BucketizedInt64ArrayPool
-	stringValuesResetFn strings.ClearArrayFn
+	boolArrayPool      *pool.BucketizedBoolArrayPool
+	intArrayPool       *pool.BucketizedIntArrayPool
+	doubleArrayPool    *pool.BucketizedFloat64ArrayPool
+	stringArrayPool    *pool.BucketizedStringArrayPool
+	int64ArrayPool     *pool.BucketizedInt64ArrayPool
+	stringArrayResetFn strings.ArrayFn
 }
 
 // NewDocsFieldBuilderOptions creates a new set of field builder options.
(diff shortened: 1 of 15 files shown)


x/pool/ref_counted_pooled_string_array_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	"github.com/xichen2020/eventdb/x/strings"
```
xichen2020: Import

index/field/docs_field.go
```diff
@@ -871,11 +872,12 @@ func (b *docsFieldBuilder) addTime(docID int32, v int64) error {
...
-	intArrayPool    *pool.BucketizedIntArrayPool
-	doubleArrayPool *pool.BucketizedFloat64ArrayPool
-	stringArrayPool *pool.BucketizedStringArrayPool
-	int64ArrayPool  *pool.BucketizedInt64ArrayPool
+	boolArrayPool       *pool.BucketizedBoolArrayPool
+	intArrayPool        *pool.BucketizedIntArrayPool
+	doubleArrayPool     *pool.BucketizedFloat64ArrayPool
+	stringArrayPool     *pool.BucketizedStringArrayPool
+	int64ArrayPool      *pool.BucketizedInt64ArrayPool
+	stringValuesResetFn strings.ClearArrayFn
```
xichen2020: Perhaps just `ArrayFn`?

index/field/docs_field.go
```diff
@@ -963,3 +965,15 @@ func (o *DocsFieldBuilderOptions) SetInt64ArrayPool(v *pool.BucketizedInt64Array
 func (o *DocsFieldBuilderOptions) Int64ArrayPool() *pool.BucketizedInt64ArrayPool {
 	return o.int64ArrayPool
 }
+
+// SetStringValuesResetFn sets a value reset function for string values.
+func (o *DocsFieldBuilderOptions) SetStringValuesResetFn(fn strings.ClearArrayFn) *DocsFieldBuilderOptions {
```
xichen2020: nit: `SetStringArrayResetFn`.

index/field/docs_field.go
```diff
@@ -963,3 +965,15 @@ func (o *DocsFieldBuilderOptions) SetInt64ArrayPool(v *pool.BucketizedInt64Array
...
+	opts := *o
+	opts.stringValuesResetFn = fn
+	return &opts
+}
+
+// StringValuesResetFn resets string values before returning a string array back to the memory pool.
+func (o *DocsFieldBuilderOptions) StringValuesResetFn() strings.ClearArrayFn {
```
xichen2020: `StringArrayResetFn`.

storage/mutable_segment.go
```diff
@@ -100,7 +101,8 @@ func newMutableSegment(
 		SetIntArrayPool(opts.IntArrayPool()).
 		SetDoubleArrayPool(opts.DoubleArrayPool()).
 		SetStringArrayPool(opts.StringArrayPool()).
-		SetInt64ArrayPool(opts.Int64ArrayPool())
+		SetInt64ArrayPool(opts.Int64ArrayPool()).
+		SetStringValuesResetFn(strings.ValuesResetFn)
```
xichen2020: `ResetArrayFn`.

xichen2020: Also add a comment here to say this is to avoid holding on to documents received after the array is returned to pool.

x/pool/ref_counted_pooled_string_array_test.go
```diff
@@ -15,7 +18,7 @@ func TestRefCountedPooledStringArray(t *testing.T) {
 	pool.Init(func(capacity int) []string { return make([]string, 0, capacity) })
 
 	vals := pool.Get(4)
-	arr := NewRefCountedPooledStringArray(vals, pool)
+	arr := NewRefCountedPooledStringArray(vals, pool, strings.ValuesResetFn)
```
xichen2020: Perhaps just leave as nil so you don't need to change the other parts of test?

x/strings/strings.go
```diff
@@ -0,0 +1,11 @@
+package strings
+
+// ClearArrayFn is the type signature of a string slice clearing fn.
+type ClearArrayFn func(values []string)
```
xichen2020: Just `ArrayFn`? More generic that way.

x/strings/strings.go
```diff
@@ -0,0 +1,11 @@
...
+
+// ValuesResetFn clears string pointers inside of a slice of strings.
+func ValuesResetFn(values []string) {
```
xichen2020: Just `ResetArray`?

x/pool/ref_counted_pooled_string_array_test.go
```diff
@@ -32,24 +33,25 @@ func TestRefCountedPooledStringArray(t *testing.T) {
 	require.Equal(t, 4, len(arr.Get()))
 	require.Equal(t, int32(2), arr2.cnt.RefCount())
 	require.Equal(t, 4, len(arr2.Get()))
-	expected1 := []string{"foo", "bar", "baz", "cat"}
-	assertReturnedToStringArrayPool(t, pool, expected1, false)
+	assertReturnedToStringArrayPool(t, pool, arr.Get(), false)
```
xichen2020: Hm I don't think you need to change the rest of the tests since you already set the reset fn to nil above. I generally prefer to assert against a very explicit value if possible so in this case I prefer to assert against an expected string array instead.

PR #132, opened 2019-02-20
Emit runtime and process metrics

black-adder:
Signed-off-by: Won Jun Jang <wjang@uber.com>

### DIFF
--- a/services/eventdb/main/main.go
+++ b/services/eventdb/main/main.go
@@ -51,6 +51,9 @@ func main() {
 	}
 	defer closer.Close()
 
+	// Report runtime and process metrics.
+	instrument.StartReportingExtendedMetrics(scope, time.Second, instrument.DetailedExtendedMetrics)
+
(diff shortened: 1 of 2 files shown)


services/eventdb/main/main.go
```diff
@@ -51,6 +51,9 @@ func main() {
 	}
 	defer closer.Close()
 
+	// Report runtime and process metrics.
+	instrument.StartReportingExtendedMetrics(scope, time.Second, instrument.DetailedExtendedMetrics)
```
xichen2020: You can configure this from the config file by adding
```
extended: moderate
```

Keep in mind that emitting detailed GC metrics will stop the world and cause GC pauses, so I'd advise using moderate if possible.

black-adder: dope

black-adder:
nvm, we already have it set... i'll add them to the graphs

PR #142, opened 2019-02-26
Add a round trip test for persist package.

notbdu:
Adds a round trip test for writing data to a file and reading fields from disk.


notbdu:
Also fixes bug: #144.

persist/fs/equals_test.go
```diff
@@ -0,0 +1,93 @@
...
+	iter2, err := f2.Values().Iter()
+	if err != nil {
+		return false, err
+	}
+	for iter1.Next() && iter2.Next() {
+		if iter1.Current() != iter2.Current() {
+			return false, nil
+		}
+	}
+	// If either iterator has extra data, they are not equal.
```
xichen2020: Need to also check `iter1.Err()` and `iter2.Err()` are nil? Same for other functions in the file.

persist/fs/reader.go
```diff
@@ -390,11 +390,12 @@ func (r *reader) mmapReadAllAndValidateChecksum(fd *os.File) ([]byte, error) {
...
 		r.logger.Warnf("warning during memory mapping info file %s: %s", fd.Name(), res.Warning)
 	}
-	if err := digest.Validate(res.Result); err != nil {
+	validatedBytes, err := digest.Validate(res.Result)
+	if err != nil {
 		mmap.Munmap(res.Result)
 		return nil, err
 	}
-	return res.Result, nil
+	return validatedBytes, nil
```
xichen2020: I think you should return a `cleanup` function to `munmap(res.Result)` in addition to `validatedBytes`, as that's only a portion of the mmap'ed region.

xichen2020:
Nice find!

persist/fs/equals_test.go
xichen2020: You can just do `require.NoError(t, iter1.Err())` for example.

PR #145, opened 2019-02-27
GRPC server part 1: write rpc endpoint

xichen2020:
cc @notbdu @cw9 @black-adder 

This PR is the part 1 of the GRPC server support. It adds the write RPC endpoint and the corresponding proto files.

Followup PRs will address:
- query RPC endpoint and corresponding proto files
- client for writing to and reading from gRPC servers.
- integration tests making use of the client to write and read data.


document/field/field.go
```diff
@@ -6,6 +6,24 @@ type Field struct {
...
+// ReturnArrayToPool returns a document array to pool.
+func ReturnArrayToPool(fields []Field, p *BucketizedFieldArrayPool) {
+	if p == nil {
+		return
+	}
+	for i := 0; i < len(fields); i++ {
+		fields[i].Reset()
+	}
+	fields = fields[:0]
+	p.Put(fields, cap(fields))
```
cw9: nit: is it better to change the Put method to just take in fields? so the user don't need to decide whether to use `cap` or `len` for the size, the `Put` method can call `cap` itself.

server/grpc/server.go
```diff
@@ -0,0 +1,64 @@
...
+	"google.golang.org/grpc/keepalive"
+)
+
+type server struct {
+	opts    *Options
+	address string
+	server  *grpc.Server
+}
+
+// NewServer creates a new http server.
```
cw9: new `grpc` server

server/grpc/service.go
```diff
@@ -0,0 +1,220 @@
...
+
+func newServiceMetrics(
+	scope tally.Scope,
+	samplingRate float64,
+) serviceMetrics {
+	batchSizeBuckets := tally.MustMakeLinearValueBuckets(0, bucketSize, numBuckets)
+	return serviceMetrics{
+		write: instrument.NewMethodMetrics(scope, "write", samplingRate),
+		batchSizeHist: scope.Tagged(map[string]string{
+			"bucket-version": strconv.Itoa(batchSizeBucketVersion),
```
cw9: question: seems the convention is to use strconv to convert a int version to string? why not just define the version as "1"?

document/field/field.go
xichen2020: Yeah so unfortunately this is dictated by the template as the array type is a `GenericValue` in the template, and does not have a `cap` method.

server/grpc/service.go
xichen2020: We could, a bit easier to find/change with a top-level const. No strong pref either way.

server/grpc/service.go
```diff
@@ -0,0 +1,220 @@
...
+const (
+	defaultReadTimeout     = time.Minute
+	defaultWriteTimeout    = time.Minute
+	batchSizeBucketVersion = 1
+	bucketSize             = 500
+	numBuckets             = 40
+)
+
+// ServiceOptions provide a set of service options.
+type ServiceOptions struct {
```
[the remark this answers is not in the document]
xichen2020: Sure

server/grpc/service.go
```diff
@@ -0,0 +1,220 @@
...
+	nowFn   clock.NowFn
+	metrics serviceMetrics
+}
+
+// NewService creates a new service.
+func NewService(db storage.Database, opts *ServiceOptions) servicepb.EventdbServer {
+	if opts == nil {
+		opts = NewServiceOptions()
+	}
+	instrumentOpts := opts.InstrumentOptions()
```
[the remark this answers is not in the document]
xichen2020: sure

services/eventdb/main/main.go
```diff
@@ -75,23 +75,29 @@ func main() {
 	defer db.Close()
 
 	// Start up HTTP server.
```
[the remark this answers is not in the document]
xichen2020: sounds good

xichen2020:
Re @black-adder : Yes that's the plan. There will be a client in `eventdb` to perform the actual write.

Issue #146, opened 2019-02-28
Integration tests querying from disk return odd results.


notbdu:
Test querying from disk fails w/: 
```
--- FAIL: TestRawQueryNoFilterOrderByFromDisk (12.41s)
    require.go:157:
                Error Trace:    raw_query_no_filter_orderby_from_disk_test.go:132
                Error:          Not equal:
                                expected: []string{"{\"service\":\"testNamespace\",\"@timestamp\":\"2019-01-22T13:25:42-08:00\",\"st\":true,\"sid\":1,\"tt\":\"active\",\"tz\":-6,\"v\":1.5}", "{\"service\":\"testNamespace\",\"@timestamp\":\"2019-01-22T13:26:42-08:00\",\"st\":true,\"sid\":1,\"tt\":\"active\",\"tz\":-6,\"v\":1.5}", "{\"service\":\"testNamespace\",\"@timestamp\":\"2019-01-22T13:27:42-08:00\",\"st\":true,\"sid\":1,\"tt\":\"active\",\"tz\":-6,\"v\":1.5}", "{\"service\":\"tes
tNamespace\",\"@timestamp\":\"2019-01-22T13:28:42-08:00\",\"st\":true,\"sid\":1,\"tt\":\"active\",\"tz\":-6,\"v\":1.5}", "{\"service\":\"testNamespace\",\"@timestamp\":\"2019-01-22T13:29:42-08:00\",\"st\":true,\"sid\":1,\"tt\":\"active\",\"tz\":-6,\"v\":1.5}", "{\"service\":\"testNamespace\",\"@timestamp\":\"2019-01-22T13:30:42-08:00\",\"st\":true,\"sid\":2,\"tt\":\"active\",\"tz\":-6,\"v\":1.5}", "{\"service\":\"testNamespace\",\"@timestamp\":\"2019-01-22T13:31:42-08:00\",\"st\":true,
[...]

PR #150, opened 2019-02-28
Add integration test for querying from disk and fix bug.

notbdu:
Adds an integration test for querying raw data from disk.

Fixes a bug where reading results from disk resulted in duplicate results. #146


values/decoding/raw_size_string_iterator.go
```diff
@@ -53,7 +52,7 @@ func (it *rawSizeStringIterator) Next() bool {
 		return false
 	}
 
-	it.curr = unsafe.ToString(it.buf[:rawSizeBytes])
+	it.curr = string(it.buf[:rawSizeBytes])
```
xichen2020: `safe.ToString`

x/hash/hash.go
```diff
@@ -22,7 +22,7 @@ func BytesHash(d []byte) Hash {
 
 // StringHash returns the hash of a string.
 func StringHash(s string) Hash {
-	return BytesHash(unsafe.ToBytes(s))
+	return BytesHash(safe.ToBytes(s))
```
xichen2020: Given the usage here is confined with a single line within a single function, and the converted bytes are not used outside the function at all, it should be fine to use `unsafe` in this simple context.

xichen2020:
Seems ci is failing due to permission error, you probably should use the temp directory as the root directory for db data, but not sure if that's directly related

notbdu:
That's weird... wonder how it was creating the data directory before...

notbdu:
Oh jk, it wasn't flushing data to disk in the other integration tests before haha.

PR #153, opened 2019-03-03
Fix premature field closing

xichen2020:
cc @notbdu 

This PR fixes the issue of closing query field too early and not closing field to filter against.


storage/segment_query.go
```diff
@@ -129,7 +129,7 @@ func applyFilter(
 	if err != nil {
 		return nil, err
 	}
-	defer fld.Close()
+	defer toFilter.Close()
```
notbdu: Ah, so it's safe to close `toFilter` after calling `toFilter.Filter` since it creates an iterator that holds references to the underlying data?

xichen2020: Actually it's because `queryFields` already holds reference to the underlying data and won't be closed until the query execution is done, so the additional reference `toFilter` holds is not necessary so it can be closed early.

PR #156, opened 2019-03-06
Use bytes instead of string internally.

notbdu:
Switch from using strings to using bytes all around internally.


generated/proto/encodingpb/string.pb.go
```diff
@@ -14,18 +14,54 @@ var _ = proto.Marshal
...
-	NumValues   int32           `protobuf:"varint,3,opt,name=num_values,json=numValues,proto3" json:"num_values,omitempty"`
-	MinValue    string          `protobuf:"bytes,4,opt,name=min_value,json=minValue,proto3" json:"min_value,omitempty"`
-	MaxValue    string          `protobuf:"bytes,5,opt,name=max_value,json=maxValue,proto3" json:"max_value,omitempty"`
+	Encoding             EncodingType    `protobuf:"varint,1,opt,name=encoding,proto3,enum=encodingpb.EncodingType" json:"encoding,omitempty"`
+	Compression          CompressionType `protobuf:"varint,2,opt,name=compression,proto3,enum=encodingpb.CompressionType" json:"compression,omitempty"`
+	NumValues            int32           `protobuf:"varint,3,opt,name=num_values,json=numValues,proto3" json:"num_values,omitempty"`
+	MinValue             []byte          `protobuf:"bytes,4,opt,name=min_value,json=minValue,proto3" json:"min_value,omitempty"`
+	MaxValue             []byte          `protobuf:"bytes,5,opt,name=max_value,json=maxValue,proto3" json:"max_value,omitempty"`
+	XXX_NoUnkeyedLiteral struct{}        `json:"-"`
```
cw9: Are we using different protobuf generator? can we unify this somehow so these files don't always get updated when different people runs proto-gen?

storage/mutable_segment.go
```diff
@@ -574,7 +573,7 @@ func (s *mutableSeg) writeTimestampFieldWithLock(docID int32, val int64) {
 func (s *mutableSeg) writeRawDocSourceFieldWithLock(docID int32, val []byte) {
 	v := field.ValueUnion{
 		Type:      field.StringType,
-		StringVal: safe.ToString(val),
+		StringVal: val,
```
cw9: do we want to switch `StringVal` to type []byte as well?

x/bytes/compare.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// GreaterThan returns true if v1 > v2.
+func GreaterThan(v1, v2 []byte) bool {
+	return bytes.Compare(v1, v2) == compareResultGreaterThan
+}
+
+// GreaterThanOrEqual returns true if v1 > v2 || v1 == v2.
+func GreaterThanOrEqual(v1, v2 []byte) bool {
+	return GreaterThan(v1, v2) || bytes.Equal(v1, v2)
```
cw9: `return -1 != bytes.Compare(v1, v2)` to save the extra `bytes.Equal` call

x/strings/strings.go
```diff
@@ -1,11 +1,11 @@
 package strings
 
 // ArrayFn is the type signature of a string slice clearing fn.
-type ArrayFn func(values []string)
+type ArrayFn func(values [][]byte)
```
cw9: Rename this file?

x/bytes/compare.go
```diff
@@ -0,0 +1,74 @@
...
+	offset := len(v1) - len(v2)
+	for idx := range v2 {
+		if v1[idx+offset] != v2[idx] {
+			return false
+		}
+	}
+	return true
+}
+
+// Equal thinly wraps `bytes.Equal` to allow for consistent imports.
```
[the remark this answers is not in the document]
notbdu: I'll just alias the import to `xbytes` then. If I move them to a `compare` package then I'll have the prefix or suffix all calls w/ `Bytes` to be explicit.

parser/json/value/value.go
```diff
@@ -128,12 +130,12 @@ func (v *Value) String() (string, error) {
 }
 
 // MustString returns a string value or panics if the value type is not string.
-func (v *Value) MustString() string {
+func (v *Value) MustString() []byte {
```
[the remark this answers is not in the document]
notbdu: I feel like we should leave this as `MustString` since we're consistent w/ naming everything `String` w/ the assumption that the underlying data is `[]byte`. If we rename this, I feel like we'd need to rename everything to be consistent.

calculation/value.go
```diff
@@ -86,7 +88,7 @@ func NewNumberUnion(v float64) ValueUnion {
 }
 
 // NewStringUnion creates a new string union.
-func NewStringUnion(v string) ValueUnion {
+func NewStringUnion(v []byte) ValueUnion {
```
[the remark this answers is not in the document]
notbdu: Same comment as `MustString` w.r.t. naming.

storage/mutable_segment.go
notbdu: It's already converted to `[]byte` in the struct def

x/strings/strings.go
notbdu: I'm going to rename the file to `bytes.go` but not the package since the naming follows the assumption we are making across the board that `strings` are `[]byte`s.

generated/proto/encodingpb/string.pb.go
notbdu: Maybe this is a versioning issue? I agree though. I'm using the same generators defined in the make command.

parser/json/value/value.go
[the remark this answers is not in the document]
cw9: I think we should rename everything to `Bytes` when the data is `bytes`, other wise there will be a big confusion in the future when you call a `getString()` but got `bytes` instead. The reason these functions and variables were named `xxxString` is because they were strings and now they are not, so the name should reflect the change as well.

storage/mutable_segment.go
cw9: sorry I meant to change the field to `bytesVal`

query/raw_result_test.go
```diff
@@ -86,7 +86,8 @@ func TestUnorderedRawResultsMarshalJSON(t *testing.T) {
 
 	b, err := json.Marshal(input)
 	require.NoError(t, err)
-	expected := `{"raw":["foo","bar","baz"]}`
+	// NB(bodu): Values are `[]byte`s in the form of base64 encoded strings.
+	expected := `{"raw":["Zm9v","YmFy","YmF6"]}`
```
cw9: nit: it's more readable to do something like `fmt.Sprintf("\"raw\":[\"%s\"]", []byte{"foo"}`?

filter/bytes_filter.go
```diff
@@ -0,0 +1,92 @@
...
+	Match(v iterator.Bytes) bool
+}
+
+type bytesToBytesFilterFn func([]byte) bytesFilterFn
+
+type bytesFilterFn func(v []byte) bool
+
+func (fn bytesFilterFn) Match(v iterator.Bytes) bool { return fn(v.Data) }
+
+func equalsBytesBytes(rhs []byte) bytesFilterFn {
```
cw9: nit: rename to `equalBytesFn`

filter/bytes_filter.go
```diff
@@ -0,0 +1,92 @@
...
+	return func(lhs []byte) bool {
+		return bytes.Equal(lhs, rhs)
+	}
+}
+
+func notEqualsBytesBytes(rhs []byte) bytesFilterFn {
```
cw9: nit: rename to `notEqualsBytesFn`

filter/bytes_filter.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+
+func smallerThanOrEqualBytesBytes(rhs []byte) bytesFilterFn {
+	return func(lhs []byte) bool {
+		return xbytes.LessThanOrEqual(lhs, rhs)
+	}
+}
+
+func startsWithBytesBytes(rhs []byte) bytesFilterFn {
```
cw9: rename to something like `hasPrefix`

filter/bytes_filter.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+
+func doesNotStartWithBytesBytes(rhs []byte) bytesFilterFn {
+	return func(lhs []byte) bool {
+		return !bytes.HasPrefix(lhs, rhs)
+	}
+}
+
+func endsWithBytesBytes(rhs []byte) bytesFilterFn {
```
cw9: rename to something like `hasSuffix`

index/field/docs_field.go
```diff
@@ -149,7 +149,7 @@ type docsField struct {
 	bf     CloseableBoolField
 	intf   CloseableIntField
 	df     CloseableDoubleField
-	sf     CloseableStringField
+	sf     CloseableBytesField
```
cw9: rename to `bf`

index/field/docs_field.go
```diff
@@ -598,7 +598,7 @@ type docsFieldBuilder struct {
 	bfb    boolFieldBuilder
 	ifb    intFieldBuilder
 	dfb    doubleFieldBuilder
-	sfb    stringFieldBuilder
+	sfb    bytesFieldBuilder
```
cw9: bfb

integration/util.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/xichen2020/eventdb/parser/json/value"
 	"github.com/xichen2020/eventdb/query"
 	xtime "github.com/xichen2020/eventdb/x/time"
+	"github.com/xichen2020/eventdb/x/unsafe"
```
cw9: should use safe?

parser/json/parser.go
```diff
@@ -9,13 +9,12 @@ import (
...
 
 	"github.com/xichen2020/eventdb/parser/json/value"
-	"github.com/xichen2020/eventdb/x/safe"
 )
 
 const (
-	trueString  = "true"
-	falseString = "false"
-	nullString  = "null"
+	trueBytes  = "true"
```
cw9: nit: this should be `trueString` lol

persist/fs/reader.go
```diff
@@ -65,7 +65,7 @@ type reader struct {
 	bd     decoding.BoolDecoder
 	id     decoding.IntDecoder
 	dd     decoding.DoubleDecoder
-	sd     decoding.StringDecoder
+	sd     decoding.BytesDecoder
```
cw9: bd

server/http/handlers/options.go
```diff
@@ -5,7 +5,7 @@ import (
 
 	"github.com/xichen2020/eventdb/parser/json"
 	"github.com/xichen2020/eventdb/parser/json/value"
-	"github.com/xichen2020/eventdb/x/safe"
+	"github.com/xichen2020/eventdb/x/unsafe"
```
cw9: use safe?

filter/bytes_filter.go
notbdu: This is consistent w/ the type specific naming of all filter fns.

```
boolToBoolFilterOps = map[Op]boolToBoolFilterFn{
        Equals:    equalsBoolBool,
        NotEquals: notEqualsBoolBool,
}
intToIntFilterOps = map[Op]intToIntFilterFn{
        Equals:             equalsIntInt,
        NotEquals:          notEqualsIntInt,
        LargerThan:         largerThanIntInt,
        LargerThanOrEqual:  largerThanOrEqualIntInt,
        SmallerThan:        smallerThanIntInt,
        SmallerThanOrEqual: smallerThanOrEqualIntInt,
}
doubleToIntFilterOps = map[Op]doubleToIntFilterFn{
        Equals:             equalsIntDouble,
        NotEquals:          notEqualsIntDouble,
        LargerThan:         largerThanIntDouble,
        LargerThanOrEqual:  largerThanOrEqualIntDouble,
        SmallerThan:        smallerThanIntDouble,
        SmallerThanOrEqual: smallerThanOrEqualIntDouble,
}
doubleToDoubleFilterOps = map[Op]doubleToDoubleFilterFn{
        Equals:             equalsDoubleDouble,
        NotEquals:          notEqualsDoubleDouble,
        LargerThan:         largerThanDoubleDouble,
        LargerThanOrEqual:  largerThanOrEqualDoubleDouble,
        SmallerThan:        smallerThanDoubleDouble,
        SmallerThanOrEqual: smallerThanOrEqualDoubleDouble,
}
intToDoubleFilterOps = map[Op]intToDoubleFilterFn{
        Equals:             equalsDoubleInt,
        NotEquals:          notEqualsDoubleInt,
        LargerThan:         largerThanDoubleInt,
        LargerThanOrEqual:  largerThanOrEqualDoubleInt,
        SmallerThan:        smallerThanDoubleInt,
        SmallerThanOrEqual: smallerThanOrEqualDoubleInt,
}
bytesToBytesFilterOps = map[Op]bytesToBytesFilterFn{
        Equals:             equalsBytesBytes,
        NotEquals:          notEqualsBytesBytes,
        LargerThan:         largerThanBytesBytes,
        LargerThanOrEqual:  largerThanOrEqualBytesBytes,
        SmallerThan:        smallerThanBytesBytes,
        SmallerThanOrEqual: smallerThanOrEqualBytesBytes,
        StartsWith:         startsWithBytesBytes,
        DoesNotStartWith:   doesNotStartWithBytesBytes,
        EndsWith:           endsWithBytesBytes,
        DoesNotEndWith:     doesNotEndWithBytesBytes,
        Contains:           containsBytesBytes,
        DoesNotContain:     doesNotContainBytesBytes,
}
timeToTimeFilterOps = map[Op]timeToTimeFilterFn{
        Equals:             equalsTimeTime,
        NotEquals:          notEqualsTimeTime,
        LargerThan:         largerThanTimeTime,
        LargerThanOrEqual:  largerThanOrEqualTimeTime,
        SmallerThan:        smallerThanTimeTime,
        SmallerThanOrEqual: smallerThanOrEqualTimeTime,
}
```

filter/bytes_filter.go
notbdu: See earlier comment on naming convention.

filter/bytes_filter.go
notbdu: See earlier comment on naming convention.

filter/bytes_filter.go
notbdu: See earlier comment on naming convention.

index/field/docs_field.go
notbdu: There is already a `bf` for the bool types.

index/field/docs_field.go
notbdu: There is already a `bfb` for bool types.

integration/util.go
notbdu: This is fine as it's usage is as follows and is self contained.

```
func parseTimestamp(v *value.Value, timestampFormat string) (int64, error) {
       str, err := v.String()
       b, err := v.Bytes()
       if err != nil {
               return 0, err
       }
       t, err := time.Parse(timestampFormat, str)
       t, err := time.Parse(timestampFormat, unsafe.ToString(b))
       if err != nil {
               return 0, err
       }
       return t.UnixNano(), nil
}
```

persist/fs/reader.go
notbdu: There is already a `bd` for bool types.

server/http/handlers/options.go
notbdu: The usage of unsafe here is self contained as well.

```
 // defaultTimeNanosFn parses the time value as a string in RFC3339 format.
 func defaultTimeNanosFn(v *value.Value) (int64, error) {
-       str, err := v.String()
+       b, err := v.Bytes()
        if err != nil {
                return 0, err
        }
-       t, err := time.Parse(time.RFC3339, str)
+       t, err := time.Parse(time.RFC3339, unsafe.ToString(b))
        if err != nil {
                return 0, err
        }
```

integration/util.go
xichen2020: Agreed that the usage is self contained but let's still use `safe` here. We should restrict usage of `unsafe` to a bare minimum (guideline being we should not use it unless the usage is simple, self-contained, and absolutely critical to performance).

server/http/handlers/options.go
xichen2020: Agreed that the usage is self contained but let's still use safe here. We should restrict usage of unsafe to a bare minimum (guideline being we should not use it unless the usage is simple, self-contained, and absolutely critical to performance).

index/field/field_iterator.go
```diff
@@ -58,12 +59,12 @@ type DoubleFieldIterator interface {
...
 
-// StringFieldIterator iterates over (doc ID, string value) pairs in a string field.
-type StringFieldIterator interface {
+// BytesFieldIterator iterates over (doc ID, string value) pairs in a string field.
+type BytesFieldIterator interface {
 	BaseFieldIterator
 
 	// Value returns the current string value.
-	Value() string
+	Value() iterator.Bytes
```
xichen2020: The `Bytes` type should probably not live in `iterator` package because it's generic.

integration/util.go
```diff
@@ -65,21 +66,32 @@ func newDocumentFromRaw(
...
 	}
-	t, err := time.Parse(timestampFormat, str)
+	t, err := time.Parse(timestampFormat, safe.ToString(b))
 	if err != nil {
 		return 0, err
 	}
 	return t.UnixNano(), nil
 }
 
+func convertRawResultToStrings(result [][]byte) []string {
```
xichen2020: I wonder if we could just compare bytes arrays directly.

integration/util.go
```diff
@@ -65,21 +66,32 @@ func newDocumentFromRaw(
...
+		strResult[idx] = string(r)
+	}
+	return strResult
+}
+
 func b(str string) []byte                                       { return []byte(str) }
 func pInt(v int) *int                                           { return &v }
 func pInt64(v int64) *int64                                     { return &v }
-func pString(v string) *string                                  { return &v }
+func pBytes(v string) *string                                   { return &v }
```
xichen2020: Hm this should still be `pString`.

parser/json/parser.go
```diff
@@ -363,26 +362,26 @@ func (p *parser) parseArray() (*value.Value, error) {
 	}
 }
 
-func (p *parser) parseString() (*value.Value, error) {
-	str, err := p.parseStringAsRaw()
+func (p *parser) parseBytes() (*value.Value, error) {
```
xichen2020: I think `parseStringAsBytes` is more accurate.

parser/json/value/iterator.go
```diff
@@ -46,10 +46,10 @@ func (it *jsonIterator) Next() bool {
 				it.value.Type = field.BoolType
 				it.value.BoolVal = v.MustBool()
 				return true
-			case StringType:
+			case BytesType:
```
xichen2020: For JSON specifically we might want to keep this as `StringType`?

parser/json/value/type.go
```diff
@@ -8,7 +8,7 @@ const (
 	UnknownType Type = iota
 	NullType
 	BoolType
-	StringType
+	BytesType
```
xichen2020: Probably should be `StringType` still for JSON values, and convert them at callsite.

notbdu:
LGTM

PR #157, opened 2019-03-12
Query immutable segments

xichen2020:
cc @black-adder @cw9 @notbdu 

This PR decouples segment implementation, storage implementation, and query execution for better separation of concerns. Additionally it changes the ingestion logic to convert each batch of writes as a new segment, and only queries immutable segments during query execution to completely remove contention during reads and writes.


index/field/docs_field.go
```diff
@@ -383,28 +388,49 @@ func (f *docsField) ShallowCopy() DocsField {
...
+		sf = osf.ShallowCopy()
 	}
-	if tf, exists := other.closeableTimeField(); exists {
-		f.setTimeField(tf.ShallowCopy())
+	if f.tf != nil {
+		tf = f.tf.ShallowCopy()
+	} else if otf, exists := other.closeableTimeField(); exists {
+		tf = otf.ShallowCopy()
 	}
+	return NewDocsField(f.fieldPath, f.fieldTypes, nf, bf, intf, df, sf, tf)
```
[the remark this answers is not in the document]
xichen2020: Yeah can do.

index/segment/segment.go
```diff
@@ -0,0 +1,200 @@
...
+// segment is a refcounted index segment.
+// The data in the segment fields may change, but the field metadata in the field map
+// does not change until the segment is closed, in which case the map is nil'ed out.
+// The segment is not thread-safe. Concurrently access to the segment (e.g., reading
+// a segment, clearing a segment, closing a segment, etc.) must be protected with
+// synchronization primitives.
+type segment struct {
+	refCountedBase
+
+	timestampFieldPath    []string
```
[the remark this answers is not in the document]
xichen2020: Meh, I slightly prefer to have them as internal fields.

index/segment/segment.go
```diff
@@ -0,0 +1,200 @@
...
+// The incoming docs field remains valid after the merge and should be closed where appropriate.
+func (f *Field) Insert(df indexfield.DocsField) {
+	f.fieldLock.Lock()
+	defer f.fieldLock.Unlock()
+
+	if f.field == nil {
+		f.field = df.ShallowCopy()
+		return
+	}
+	merged := f.field.NewMergedDocsField(df)
```
notbdu: What's the case where a field would be merged? I ony see `Insert()` being called by the `updatingSegmentRetriever` when `lookupFields` fails to find a `DocsField()` on a segment. Wouldn't this mean that either a field is either completely in memory or needs to be wholly retrieved from disk?

I can see a merge potentially being triggered by concurrent retrievals of the same field from disk. Is that the intent? In that case, wouldn't we not want to load the field into memory at all if it's already being loaded from memory but not yet `Insert`ed?

storage/namespace.go
```diff
@@ -101,31 +128,70 @@ func newDatabaseNamespace(
...
-	doc document.Document,
+	docs []document.Document,
 ) error {
-	shard, err := n.shardFor(doc.ID)
-	if err != nil {
-		return err
+	builder := n.newSegmentBuilderFn(n.segmentBuilderOpts)
+	builder.AddBatch(docs)
+	rawSegment := builder.Build()
+	newSegment := newDatabaseSegment(n.id, rawSegment, n.segmentOpts)
```
notbdu: Do we want to create a new segment for every batch? Some batches could be very small. OTOH, I'm guessing this is to make it so that we're only querying immutable segments?

Is the intention that we move immediately to adding compaction to the flushing process?

I also don't see `MaxNumDocsPerSegment` used anywhere.

index/segment/segment.go
xichen2020: Not necessarily. Individual types of fields can be retrieved separately. IOW you can have a multi-typed fields and you have some of the typed values in memory whereas the values for other types on disk.

storage/namespace.go
xichen2020: Yes. Agreed on potential small batch sizes. We can either do more aggressive batching on the ingestion tier, or having some kind of ingestion queue in the database that batches things up on a fixed time interval. Either way this mechanism can be extended to handle that.

If ingestion teir handles batching logs into large batches, we don't need compaction immediately. Also I'd be curious to see what the query perf looks like with no compaction. But intuitively I think we'll need compaction at some point.

Not sure why linter doesn't complain about `MaxNumDocsPerSegment`. I'll remove.

PR #158, opened 2019-03-16
Eagerly flush segments on batched writes

xichen2020:
cc @black-adder @cw9 @notbdu 

This PR adds logic to eagerly flush segments on each batched write so we don't hold on to the segments in memory to avoid blowing up memory usage. It is configurable whether the `WriteBatch` call will return a response synchronously (i.e., only ack after the segment is flushed) or asynchronously (i.e., ack as soon as the segment is queued for flush).

Also fixed a number of bugs with encoding/decoding logic:
- Typed fields don't have type name in file names causing one type to override another type
- Types with bit encoded values (`int` and `time`) add padding during encoding causing incorrect number of values to be decoded
- If int value is dictionary encoded, filtering does not work properly due to incorrectly using `varintIterator` to decode indices even though they are bit encoded (and hence should use a bit-based int decoder).
- Filtering does not work for values with multiple types (e.g., int filter doesn't work w/ double values and vice versa).


storage/flush_mgr.go
```diff
@@ -0,0 +1,202 @@
...
+	m.RLock()
+	defer m.RUnlock()
+
+	if m.state != flushManagerOpen {
+		return errFlushManagerNotOpenOrClosed
+	}
+	select {
+	case m.unflushed <- p:
+	default:
+		return errFlushManagerFlushQueueFull
```
black-adder: feel like this deserves its own metric

storage/namespace.go
```diff
@@ -172,21 +165,63 @@ func (n *dbNamespace) WriteBatch(
...
+
+		// On flush success, clear in-memory data (but keep the metadata) and update the list
+		// of searchable segments.
+		rawSegment.ClearAll()
+
+		n.Lock()
+		defer n.Unlock()
+
+		if n.closed {
+			rawSegment.Close()
```
black-adder: hmm, we'll have an odd situation where the segment has already been persisted to disk but not searchable.

storage/fs_mgr.go
```diff
@@ -0,0 +1,57 @@
...
+func newFileSystemManager(database database, opts *Options) *fileSystemManager {
+	return &fileSystemManager{
+		database: database,
+		opts:     opts,
+		logger:   opts.InstrumentOptions().Logger(),
+		nowFn:    opts.ClockOptions().NowFn(),
+	}
+}
+
+func (mgr *fileSystemManager) Run() bool {
```
black-adder: if this manager is only going to handle compaction, maybe call it compactionManager?

values/decoding/bit_stream_int_iterator.go
```diff
@@ -0,0 +1,73 @@
...
+	numEncodedValues    int
+
+	curr  int
+	count int
+	err   error
+}
+
+func newBitStreamIntIterator(
+	reader xio.Reader,
+	bitsPerEncodedValue int, // Number of bits per encoded value
```
black-adder: the variable names are pretty self-explanatory, comments don't seem necessary.

values/decoding/delta_int_iterator.gen.go
```diff
@@ -5,6 +5,8 @@
 package decoding
 
 import (
+	"fmt"
+
```
black-adder: rm

values/decoding/delta_time_iterator.gen.go
```diff
@@ -5,6 +5,8 @@
 package decoding
 
 import (
+	"fmt"
+
```
black-adder: rm

values/decoding/fs_bool_values.go
```diff
@@ -47,7 +49,7 @@ func (v *fsBasedBoolValues) Filter(
 		return nil, errNilFilterValue
 	}
 	if filterValue.Type != field.BoolType {
-		return nil, errUnexpectedFilterValueType
+		return nil, fmt.Errorf("bool values filter expect bool filter value type but got %v filter value type", filterValue.Type)
```
black-adder: s/expect/expects

storage/namespace.go
```diff
@@ -172,21 +165,63 @@ func (n *dbNamespace) WriteBatch(
...
+			return errNamespaceAlreadyClosed
+		}
+		newSegment := newDatabaseSegment(n.id, rawSegment, n.segmentOpts)
+		n.sealedByMaxTimeAsc.Set(float64(rawSegment.Metadata().MaxTimeNanos), newSegment)
+		return nil
+	}
+
+	if n.opts.WriteAsync() {
+		go callbackFn()
```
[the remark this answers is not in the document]
xichen2020: Yeah possibly, although for the PoC the rate of writes coming in is probably low enough. Also we by default disable async writes.

storage/flush_mgr.go
xichen2020: Makes sense. Will add.

storage/namespace.go
xichen2020: Yeah this is ok because it's done during namespace close (i.e., database shutdown), so the database will find such segments when the db starts up with proper bootstrapping logic.

storage/fs_mgr.go
xichen2020: There will also be file cleanup tasks (e.g., to delete expired data).

values/decoding/bit_stream_int_iterator.go
xichen2020: Sure can remove.

values/decoding/delta_int_iterator.gen.go
xichen2020: This is also generated.

values/decoding/delta_time_iterator.gen.go
xichen2020: Same, this is auto generated.

